Issue #2, opened 2020-05-14
Automatic Throttling & Retrying


XAMPPRocky:
Currently it up to the user to properly rate limit their application, it would be nice to provide the option to have `octocrab` handle that.

PR #5, opened 2020-05-15
Proofread README.md

LeSeulArtichaut:
Fixes some typos. I'm not a native English speaker, so I may be wrong for some of those.


XAMPPRocky:

LeSeulArtichaut:
> congrats on your first contribution! 🎉

It won't be the last I hope :D
I might take a look at #2, but that is a challenge I may not be able to overcome 😄

XAMPPRocky:
I think how you would do that is implement the retry logic in `Octocrab::execute`. You would need to add the rate limiting configuration and variables in `Octocrab`. They'd need to be stored in some kind of thread and atomic wrapper.

https://developer.github.com/v3/#rate-limiting

PR #6, opened 2020-05-15
Add the `Octocrab::current_user` method

LeSeulArtichaut:
I think this will be quite useful, from what I'd need to do in [rust-lang/triagebot](https://github.com/rust-lang/triagebot).
@XAMPPRocky please comment on the design, I'm not completely sure about how to handle unauthenticated users. Maybe I can avoid doing a request if the auth token is not set?

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -3,3 +3,4 @@ pub mod markdown;
 pub mod gitignore;
 pub mod orgs;
 pub mod pulls;
+pub mod current;
--- a/src/api/current.rs
+++ b/src/api/current.rs
@@ -1,0 +1,16 @@
+use crate::{models, Octocrab, Result};
+
+pub struct CurrentAuthHandler<'octo> {
+    crab: &'octo Octocrab,
+}
+
+impl<'octo> CurrentAuthHandler<'octo> {
+    pub(crate) fn new(crab: &'octo Octocrab) -> Self {
+        Self { crab }
+    }
+
+    /// Fetches information about the current user.
+    pub async fn user(&self) -> Result<models::User> {
+        self.crab.get("/user", None::<&()>).await
+    }
+}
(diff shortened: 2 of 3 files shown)


XAMPPRocky:
Thank you for your PR! However I had a slightly different design in mind for this method, as there's actually a lot of methods that involve the "current" user. I would accept if this was similar `issues` where there's a `CurrentAuthHandler` (or similar name) that has a `user` method. So you would then call it like the following.

```rust
let user = octocrab:instance().current().user().await?;
```

LeSeulArtichaut:
@XAMPPRocky I reorganized the API, can you confirm this is what you had in mind?

src/api/users.rs
```diff
@@ -0,0 +1,30 @@
...
+}
+
+impl<'octo> CurrentAuthHandler<'octo> {
+    pub(crate) fn new(crab: &'octo Octocrab) -> Self {
+        Self { crab }
+    }
+
+    /// Fetches information about the current user.
+    /// If the current client is unauthorized, this returns an `Ok(None)` value.
+    pub async fn user(&self) -> Result<Option<models::User>> {
```
XAMPPRocky, line 14: I'd remove the error path, and change the return type to `Result<models::User>`. I'd consider it programmer error to use the `CurrentAuthHandler` when using a token that doesn't have an authenticated user behind. And people who need to handle it will get an error anyway.

LeSeulArtichaut:
@XAMPPRocky Resolved your review and the merge conflicts.

src/api.rs
```diff
@@ -3,3 +3,4 @@ pub mod markdown;
 pub mod gitignore;
 pub mod orgs;
 pub mod pulls;
+pub mod users;
```
XAMPPRocky, line 6:
```suggestion
pub mod current;
```

LeSeulArtichaut:
@XAMPPRocky This should be ready to go then! I had a question though, should I re-export the `current` module just like all the other API modules?

XAMPPRocky:
Thank you again for your PR! This good to me.

PR #7, opened 2020-05-16
(Try to) implement rate limiting

### LINKED ISSUE #2: Automatic Throttling & Retrying

LeSeulArtichaut:
I have absolutely no experience in implementing rate limiting mechanisms, so this approach is probably very naive. Also, the search API has different limits, which I don't take into account yet.
I haven't had time to test this either, but I wanted to first ask if the code seems right and if the approach seems viable.

Closes #2.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,3 +33,4 @@ hyperx = "1.0.0"
 snafu = { version = "0.6.6", features = ["backtraces"] }
 once_cell = "1.3.1"
 arc-swap = "0.4.6"
+cfg-if = "0.1.10"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,11 @@ documentation = "https://docs.rs/octocrab"
 # and only builds that target
 targets = ["x86_64-unknown-linux-gnu"]
 
+[features]
+default = []
+
+rate-limiting = []
+
 [dependencies]
 reqwest = { version = "0.10.4", features = ["json"] }
 tokio = { version = "0.2.17", features = ["full"] }
--- a/src/api/meta.rs
+++ b/src/api/meta.rs
@@ -1,0 +1,23 @@
+use crate::{models, Octocrab, Result};
+
+pub struct MetaHandler<'octo> {
+    crab: &'octo Octocrab,
+}
+
+impl<'octo> MetaHandler<'octo> {
+    pub(crate) fn new(crab: &'octo Octocrab) -> Self {
+        Self { crab }
+    }
+
+    /// Fetches your current rate limit status.
+    /// ```no_run
+    /// # async def run() -> octocrab::Result<()> {
+    /// # let octocrab = octocrab::Octocrab::default();
+    /// octocrab.meta().rate_limits().await?;
+    /// # Ok(())
+    /// # }
+    pub async fn rate_limits(&self) -> Result<models::ResourcesRateLimits> {
+        let limits: Result<models::RateLimits> = self.crab.get("/rate_limit", None::<&()>).await;
+        limits.map(|l| l.resources)
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -149,9 +149,6 @@ mod error;
 mod from_response;
 mod page;
 
-#[cfg(feature = "rate-limiting")]
-mod rate_limiting;
-
 pub mod models;
 pub mod params;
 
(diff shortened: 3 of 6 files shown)


src/lib.rs
```diff
@@ -157,6 +160,15 @@ use snafu::*;
 
 use auth::Auth;
 
+#[cfg(feature = "rate-limiting")]
```
XAMPPRocky, line 163: I would recommend importing `cfg_if` and using that to import instead of having to say `#[cfg(feature = "rate-limiting")]` each time.

src/lib.rs
```diff
@@ -482,10 +500,61 @@ impl Octocrab {
...
+
+        {
+            self.rate_limiter
+                .as_ref()
+                .lock()
+                .expect("Poisoned rate limiter mutex")
+                .register_request();
+        }
+
+        let request_cloned = request.try_clone().unwrap(); // we never use streams in bodies
```
XAMPPRocky, line 530: Instead of trying to send the request and seeing if it fails, I would instead call the `/rate_limit` endpoint and see how much is left. If you'd also like the public method for this, I would put it under `MetaHandler` struct similar how the rest the API is laid out. https://developer.github.com/v3/rate_limit/

src/rate_limiting.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    pub fn register_request(&mut self) {
+        self.current_count += 1;
+    }
+
+    pub fn register_response(&mut self, res: &crate::Result<Response>) {
+        if let Ok(ref res) = res {
+            let headers = res.headers();
+            if let (Some(remaining), Some(reset)) = (headers.get("X-RateLimit-Remaining"), headers.get("X-RateLimit-Reset")) {
+                let (remaining, reset) = (remaining.to_str().unwrap().parse().unwrap(), Utc.timestamp(reset.to_str().unwrap().parse().unwrap(), 0));
```
XAMPPRocky, line 69: I'd refactor this to collapse all the unwraps into a single `if let` check. I haven't tested this code, but it should work.

```suggestion
            let remaining = headers.get("X-RateLimit-Remaining")
                .and_then(|v| v.to_str().ok())
                .and_then(|v| v.parse().ok());
            let reset = headers.get("X-RateLimit-Reset")
                .and_then(|v| v.to_str().ok())
                .and_then(|v| v.parse().ok());
            if let (Some(remaining), Some(reset)) = (remaining, reset) {
```

Cargo.toml
```diff
@@ -16,6 +16,11 @@ documentation = "https://docs.rs/octocrab"
 # and only builds that target
 targets = ["x86_64-unknown-linux-gnu"]
 
+[features]
+default = []
+
+rate-limiting = []
```
XAMPPRocky, line 22: This doesn't have to be done until we're closer to merging, but I would like to also include some documentation in the README and the root library documentation about rate limiting.

src/lib.rs
LeSeulArtichaut: This was actually meant as a fallback, in case another source of requests hits the rate limit. In other cases, the `limiter.request_delay().await` should wait if we are rate-limited.

src/models.rs
```diff
@@ -736,3 +736,30 @@ pub struct InstallationToken {
...
 }
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RateLimits {
+    pub resources: ResourcesRateLimits,
+    #[deprecated(
+        note = "This has been deprecated by GitHub. Use `RateLimits::resources::core` instead."
+    )]
+    pub rate: RateLimit,
```
XAMPPRocky, line 747:
```suggestion
```

src/lib.rs
```diff
@@ -513,10 +535,61 @@ impl Octocrab {
...
+            limiter.register_response(&response);
+            if let RateLimiterState::RateLimited(reset) = &limiter.state {
+                (*reset - Utc::now()).to_std().ok()
+            } else {
+                None
+            }
+        };
+
+        if let (Ok(res), Some(delay)) = (&response, &delay) {
+            if res.status() == reqwest::StatusCode::FORBIDDEN {
```
XAMPPRocky, line 582: This should have the same behaviour as octokit's retry plugin, which doesn't retry on Forbidden. https://github.com/octokit/plugin-retry.js

src/api/meta.rs
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    /// Fetches your current rate limit status.
+    /// ```no_run
+    /// # async def run() -> octocrab::Result<()> {
+    /// # let octocrab = octocrab::Octocrab::default();
+    /// octocrab.meta().rate_limits().await?;
+    /// # Ok(())
+    /// # }
+    pub async fn rate_limits(&self) -> Result<models::ResourcesRateLimits> {
```
XAMPPRocky, line 19:
```suggestion
    /// octocrab.meta().get_rate_limit().await?;
    /// # Ok(())
    /// # }
    pub async fn get_rate_limit(&self) -> Result<models::ResourcesRateLimits> {
```

src/lib.rs
LeSeulArtichaut: Got it. I'll take a look.

PR #12, opened 2020-05-19
Add parts of the Repositories API

LeSeulArtichaut:
Again, needs documentation, but I wanted to make sure the API seems correct.

### DIFF
--- a/src/api/repos.rs
+++ b/src/api/repos.rs
@@ -1,0 +1,81 @@
+//! The repositories API.
+
+mod file;
+
+pub use file::UpdateFileBuilder;
+
+use crate::{models, params, Octocrab, Result};
+
+pub struct RepoHandler<'octo> {
+    crab: &'octo Octocrab,
+    owner: String,
+    repo: String,
+}
+
+impl<'octo> RepoHandler<'octo> {
+    pub(crate) fn new(crab: &'octo Octocrab, owner: String, repo: String) -> Self {
+        Self { crab, owner, repo }
+    }
+
+    pub async fn get_ref(&self, reference: &params::repos::Reference) -> Result<models::Ref> {
+        let url = format!(
+            "/repos/{owner}/{repo}/git/ref/{reference}",
+            owner = self.owner,
+            repo = self.repo,
+            reference = reference.ref_url(),
+        );
+        self.crab.get(url, None::<&()>).await
+    }
+
+    pub async fn create_ref(
+        &self,
+        reference: &params::repos::Reference,
+        sha: String,
+    ) -> Result<models::Ref> {
+        let url = format!(
+            "/repos/{owner}/{repo}/git/refs",
+            owner = self.owner,
+            repo = self.repo,
+        );
+        self.crab
+            .post(
+                url,
+                Some(&serde_json::json!({
+                    "ref": reference.ref_url(),
+                    "sha": sha,
+                })),
+            )
+            .await
+    }
+
+    pub fn create_file(
+        &self,
+        path: impl Into<String>,
+        message: impl Into<String>,
+        content: impl AsRef<[u8]>,
+    ) -> UpdateFileBuilder<'_, '_> {
+        UpdateFileBuilder::new(
+            self,
+            path.into(),
+            message.into(),
+            base64::encode(content),
+            None,
+        )
+    }
+
+    pub fn update_file(
+        &self,
+        path: impl Into<String>,
+        message: impl Into<String>,
+        content: impl AsRef<[u8]>,
+        sha: impl Into<String>,
+    ) -> UpdateFileBuilder<'_, '_> {
+        UpdateFileBuilder::new(
+            self,
+            path.into(),
+            message.into(),
+            base64::encode(content),
+            Some(sha.into()),
+        )
+    }
+}
--- a/src/api/repos.rs
+++ b/src/api/repos.rs
@@ -6,6 +6,11 @@ pub use file::UpdateFileBuilder;
 
 use crate::{models, params, Octocrab, Result};
 
+/// Handler for GitHub's repository API.
+///
+/// Created with [`Octocrab::repos`].
+///
+/// [`Octocrab::repos`]: ../struct.Octocrab.html#method.repos
 pub struct RepoHandler<'octo> {
     crab: &'octo Octocrab,
     owner: String,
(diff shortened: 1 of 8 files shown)


src/api/repos.rs
```diff
@@ -0,0 +1,173 @@
...
+            path.into(),
+            message.into(),
+            base64::encode(content),
+            Some(sha.into()),
+        )
+    }
+
+    /// Creates a new `PullRequestHandler` for the repository,
+    /// that allows you to access GitHub's pull request API.
+    pub fn pulls(&self) -> super::pulls::PullRequestHandler {
```
XAMPPRocky, line 170: Is there a reason to add this?

src/api/repos/file.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// The author of the file.
+    pub fn author(mut self, author: impl Into<models::AuthorUser>) -> Self {
+        self.author = Some(author.into());
+        self
+    }
+
+    /// Sends the actual request.
+    pub async fn send(self) -> Result<serde_json::Value> {
+        // FIXME: change return type
```
XAMPPRocky, line 59: Leaving comment here so that you can mark it as resolved when this is fixed.

src/lib.rs
```diff
@@ -467,28 +477,34 @@ impl Octocrab {
 
     /// Send a `PUT` request to `route` with optional query parameters,
     /// returning the body of the response.
-    pub async fn put<R, A, P>(&self, route: A, parameters: Option<&P>) -> Result<R>
+    pub async fn put<R, A, P, B>(&self, route: A, parameters: Option<&P>, body: Option<&B>) -> Result<R>
```
XAMPPRocky, line 480: Instead of adding an argument I would replace `parameters`, it's not being used anywhere and then it matches `post`
```suggestion
    pub async fn put<R, A, B>(&self, route: A, body: Option<&B>) -> Result<R>
```

src/lib.rs
```diff
@@ -467,28 +477,34 @@ impl Octocrab {
...
         R: FromResponse,
     {
-        let response = self._put(self.absolute_url(route)?, parameters).await?;
+        let response = self._put(self.absolute_url(route)?, parameters, body).await?;
         R::from_response(Self::map_github_error(response).await?).await
     }
 
     /// Send a `PATCH` request with no additional post-processing.
-    pub async fn _put<P: Serialize + ?Sized>(
+    pub async fn _put<P: Serialize + ?Sized, B: Serialize + ?Sized>(
```
XAMPPRocky, line 492: Same as above.

src/models.rs
```diff
@@ -736,3 +736,30 @@ pub struct InstallationToken {
     pub permissions: Permissions,
     pub repositories: Option<Vec<Repository>>,
 }
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
XAMPPRocky, line 741: Why camelCase? I'm pretty sure GitHub uses "snake_case" in their JSON.

src/params.rs
```diff
@@ -214,4 +214,23 @@ pub mod repos {
...
+
+    impl Reference {
+        pub fn ref_url(&self) -> String {
+            match self {
+                Self::Branch(branch) => format!("heads/{}", branch),
+                Self::Tag(tag) => format!("tags/{}", tag),
+            }
+        }
+
+        pub fn full_ref_url(&self) -> String {
```
XAMPPRocky, line 232: Instead of or in addition to this method, I would add a `Display` impl that's equivalent.

src/params.rs
```diff
@@ -214,4 +214,23 @@ pub mod repos {
         Pushed,
         FullName,
     }
+
+    /// A Git reference, either a branch or a tag.
+    pub enum Reference {
```
XAMPPRocky, line 219:
```suggestion
    #[derive(Debug, Clone)]
    pub enum Reference {
```

src/api/repos.rs
LeSeulArtichaut: Ease of use? I found it useful when writing my example of creating a file then opening a PR to the repository, so I added it, but I can remove it.

src/models.rs
LeSeulArtichaut: Mhh you're right. Another example of missed copy-pasting :eyes:

src/lib.rs
LeSeulArtichaut: Are you sure? It seems like it is used when locking an issue ([GitHub docs](https://developer.github.com/v3/issues/#lock-an-issue), [code in Octocrab](https://github.com/XAMPPRocky/octocrab/blob/f011c2d23ce9e54e76a42eef83c625a993e640e0/src/api/issues.rs#L160-L168))

XAMPPRocky: Yeah, my code is actually wrong, and this would fix it. Locking an issue wants you to send a JSON body not query parameters.

src/api/repos.rs
XAMPPRocky: For now I would prefer that, just to try to limit the API, and keep it similar to octokit. Once API is a lot more fleshed out, used, and stable, we can add more convenience methods.

LeSeulArtichaut:
@XAMPPRocky Do you have any procedure/tips for creating a model? Or do I just take the example in the GitHub docs and write it as a Rust struct?

LeSeulArtichaut:
(Rebased to resolve conflicts)

XAMPPRocky:
@LeSeulArtichaut If it's small enough I do it by hand, if it's large I use this website and then clean it up. https://transform.tools/json-to-rust-serde

LeSeulArtichaut:
@XAMPPRocky I tried to create the model: `FileUpdate`, which is used for both creating and updating a file. I hope I did it right, please take a close look when reviewing 🤞

src/api/repos/file.rs
```diff
@@ -0,0 +1,67 @@
+use super::*;
+
+#[derive(serde::Serialize)]
+pub struct UpdateFileBuilder<'octo, 'r> {
```
XAMPPRocky, line 4: Could you add a serialize test for this builder?

LeSeulArtichaut:
@XAMPPRocky Just added the test.

XAMPPRocky:
Thank you for your PR!

PR #18, opened 2020-05-22
Add `Page::{first, last, number_of_pages}`

### LINKED ISSUE #15: Add first, last, and number_of_pages to Page
Currently Page only has next and previous fields, but GitHub will also at times give you the first and last page in a paginated response. On top of that it would be nice if it also had a method that gave you the total number of pages if last is present.

https://developer.github.com/v3/guides/traversing-with-pagination/

LeSeulArtichaut:
Closes #15.


XAMPPRocky:
Thank you for your PR!

PR #21, opened 2020-05-23
Improves CI

Elinvynia:
~~Note: This requires the GITHUB_TOKEN secret to be set in the repo settings for the pretty clippy direct lints, it's optional but nice to have.~~

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -6,26 +6,11 @@ on:
 
 jobs:
   build:
+    runs-on: ${{ matrix.os }}-latest
     strategy:
       matrix:
-        type: [stable, beta, nightly, windows, macos]
-        include:
-          - type: stable
-            rust: stable
-            os: ubuntu-latest
-          - type: beta
-            rust: beta
-            os: ubuntu-latest
-          - type: nightly
-            rust: nightly
-            os: ubuntu-latest
-          - type: windows
-            rust: stable
-            os: windows-latest
-          - type: macos
-            rust: stable
-            os: macos-latest
-    runs-on: ${{ matrix.os }}
+        channel: [stable, beta, nightly]
+        os: [ubuntu, macos, windows]
     steps:
     - uses: actions/checkout@v2
     - run: rustup default ${{ matrix.rust }}
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -27,6 +27,7 @@ jobs:
             os: macos-latest
     runs-on: ${{ matrix.os }}
     steps:
+    - uses: actions/checkout@v2
     - run: rustup default ${{ matrix.rust }}
     - run: cargo build --verbose --all-targets
     - run: cargo test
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR! However I would prefer to not use the `actions-rs`actions. In my opinion it's too much boilerplate now compared to using rustup directly which is always pre-installed. the clippy check would be nice but it has limitation of not working on forked repositories which makes unusable in my opinion as all pull requests are from forked repositores.

If you rewrote the clippy and format checks to use cargo directly using `run:` I would be happy to merge this in.

Elinvynia:
> Thank you for your PR! However I would prefer to not use the `actions-rs`actions. In my opinion it's too much boilerplate now compared to using rustup directly which is always pre-installed. the clippy check would be nice but it has limitation of not working on forked repositories which makes unusable in my opinion as all pull requests are from forked repositores.
> 
> If you rewrote the clippy and format checks to use cargo directly using `run:` I would be happy to merge this in.

When was rustup added to the default installed packages? I definitely wasn't aware of this change, so thank you for bringing this to my attention, definitely gonna use it on my repos too.

XAMPPRocky:
@Elinvynia I believe it was added at the end of last year. You can read what's exactly available here. https://help.github.com/en/actions/reference/software-installed-on-github-hosted-runners

.github/workflows/rust.yml
```diff
@@ -6,13 +6,40 @@ on:
...
+          - type: nightly
+            rust: nightly
+            os: ubuntu-latest
+          - type: windows
+            rust: stable
+            os: windows-latest
+          - type: macos
+            rust: stable
+            os: macos-latest
+    runs-on: ${{ matrix.os }}
```
XAMPPRocky, line 28: Can you either revert the change to the strategy matrix or explain the change? It doesn't look like it adds anything to me.

Elinvynia: In fact it removes a tiny bit - I am open to doing it your way but this saves a little bit of processing power to only test beta and nightly on linux rather than all platforms, so shorter CI times and less runners used allowing this repo to scale better in the future at the cost of minor loss of information in case something goes wrong on these two platforms on the non-stable channel.

XAMPPRocky: Yeah, I want to test beta and nightly on macos and windows too. I don't care about CI times or less runners. It takes less than 5 minutes, and I'd prefer to be able catch a upcoming breaking change or regression in Rust.

Elinvynia: Alright, gonna revert it!

.github/workflows/rust.yml
```diff
@@ -13,6 +13,18 @@ jobs:
         os: [ubuntu, macos, windows]
     steps:
     - uses: actions/checkout@v2
-    - run: rustup default ${{ matrix.channel }}
+    - run: rustup default ${{ matrix.rust }}
```
XAMPPRocky, line 16:
```suggestion
    - run: rustup default ${{ matrix.channel }}
```

XAMPPRocky:
Thank you again for your PR!

PR #26, opened 2020-07-10
Add endpoint for code search

nathanwhit:
Adds an endpoint for code search, as described [here](https://docs.github.com/en/rest/reference/search#search-code).

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -694,6 +694,17 @@ pub struct Commit {
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct Code {
+    pub name: String,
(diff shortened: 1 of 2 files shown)


src/models.rs
```diff
@@ -693,6 +711,17 @@ pub struct Commit {
     pub committer: User,
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct Code {
```
XAMPPRocky, line 715:
```suggestion
#[non_exhaustive]
pub struct Code {
```

XAMPPRocky:
Thank you again for your PR, and congrats on your first contribution! 🎉

PR #29, opened 2020-07-28
Add cargo fmt to github actions

Atul9:


XAMPPRocky:
Thank you for your PR! However I don't think I want to check formatting in CI. I don't really find that it provides a lot of benefit compared to the pain of getting a CI failure solely on style. I'll occasionally run `cargo fmt` manually and I feel that works good enough for me.

PR #33, opened 2020-08-17
Edit Comments on Issues

4meta5:
I used it here and it worked https://github.com/sunshine-protocol/sunshine-bounty/pull/164 . I needed to get the last comment and update it instead of doing a new comment every time.

Thanks for making this by the way -- I appreciate your work!

### DIFF
--- a/src/api/issues.rs
+++ b/src/api/issues.rs
@@ -531,6 +531,33 @@ impl<'octo> IssueHandler<'octo> {
         self.crab.get(route, None::<&()>).await
     }
 
+    /// Creates a comment in the issue.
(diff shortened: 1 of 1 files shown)


src/api/issues.rs
```diff
@@ -531,6 +531,33 @@ impl<'octo> IssueHandler<'octo> {
         self.crab.get(route, None::<&()>).await
     }
 
+    /// Creates a comment in the issue.
```
4meta5, line 534:
```suggestion
    /// Updates a comment in the issue.
```

XAMPPRocky:

PR #57, opened 2021-01-01
Parse the boolean values on Pullrequests to default false.

Yatekii:
I added this small change, as the GH API seems to omit false values. After this change the PR::list() method works again :)


XAMPPRocky:

XAMPPRocky:
This has now been published in 0.8.5

Yatekii:
That was really fast, thank you!

PR #59, opened 2021-01-02
Make Tag struct fields public

### LINKED ISSUE #58: How to list tag names?
list_tags() returns a page of models::repos::Tag, but then Tag model contains just private fields so it is not possible to fetch tag names from there.

acim:
Resolves #58


XAMPPRocky:

XAMPPRocky:
This has now been published as 0.8.6

Issue #25, opened 2020-07-07
Add Events API


fenhl:
Support for [the Events API](https://docs.github.com/en/rest/reference/activity#events) would be nice.

wayofthepie:
I started looking at this here https://github.com/XAMPPRocky/octocrab/compare/master...wayofthepie:events-api?expand=1. I've been using an extension on `octocrab` for this API for a while now anyway, so said I might as well clean it up and try contribute it back. Lots left to do, but should have a PR soon enough.

XAMPPRocky:
Thank you for your issue! A PR is more than welcome.

wayofthepie:
Just adding this to track what's done/left to do in the main models.
- [ ] CommitCommentEvent
- [x] CreateEvent
- [ ] DeleteEvent
- [ ] ForkEvent
- [ ] GollumEvent
- [ ] IssueCommentEvent
- [x] IssuesEvent
- [ ] MemberEvent
- [ ] PublicEvent
- [ ] PullRequestEvent
- [ ] PullRequestReviewCommentEvent
- [x] PushEvent
- [ ] ReleaseEvent
- [ ] SponsorshipEvent
- [ ] WatchEvent

PR #60, opened 2021-01-03
Add initial support for event handling

wayofthepie:
Hello! This PR adds initial support for Events (#25), specifically the repo events API `/repos/{owner}/{repo}/events`. There is a bit here, quick overview of the main changes:

  * Adds direct support for `CreateEvent` and `PushEvent` (see [EventType](https://github.com/XAMPPRocky/octocrab/pull/60/files#diff-279c8cd5a7a2067dc9b53f644006e103e949cce2085d0ea28b18d5153e2554daR24-R30)) , all other events are currently mapped to a `serde_json::Value` type.
  * Adds an [etag module](https://github.com/XAMPPRocky/octocrab/pull/60/files#diff-0d0e00934e22d61e855109d06fedb297be71f02211fa836055030479c1beed4a) for handling of etags, which are really necessary to use the Events API correctly. 
  * Adds some [high level tests](https://github.com/XAMPPRocky/octocrab/pull/60/files#diff-6a98530cb09e3005dbdc349918639b85df67aa68744d38d2392835efc26810e4) for the repo events calls using [wiremock](https://docs.rs/wiremock/0.4.3/wiremock/). These were useful to get the etag handling right, but not sure if you want these here as none exist so far.

I plan on adding the models for the rest of the events, and the other events API's over the next few weeks. But thought I'd raise this first as it is usable now and you may want to change a few things in how I've implemented it. Thanks!

### DIFF
--- a/src/etag.rs
+++ b/src/etag.rs
@@ -1,7 +1,5 @@
 //! Types for handling etags.
-use reqwest::header::HeaderValue;
-use snafu::ResultExt;
-use std::convert::TryFrom;
+pub use hyperx::header::EntityTag;
 
 /// Represents resources identified by etags.
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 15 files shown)


src/models/events.rs
```diff
@@ -0,0 +1,294 @@
...
+use reqwest::Url;
+use serde::{de::Error, Deserialize, Serialize};
+
+/// A GitHub event.
+#[derive(Debug, Clone, PartialEq, Serialize)]
+#[non_exhaustive]
+pub struct Event {
+    pub id: String,
+    #[serde(rename = "type")]
+    pub typ: EventType,
```
XAMPPRocky, line 14: I find using the exact name as the JSON response more intuitive, even if you have to use the rare `r#` syntax.
```suggestion
    pub r#type: EventType,
```

src/models/events/payload.rs
```diff
@@ -0,0 +1,55 @@
...
+
+/// The payload in a [`EventPayload::PushEvent`] type.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[non_exhaustive]
+pub struct PushEventPayload {
+    pub push_id: u64,
+    pub size: u64,
+    pub distinct_size: u64,
+    #[serde(rename = "ref")]
+    pub ref_field: String,
```
XAMPPRocky, line 27:
```suggestion
    pub r#ref: String,
```

src/models/events/payload.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub commits: Vec<Commit>,
+}
+
+/// The payload in a [`EventPayload::CreateEvent`] type.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[non_exhaustive]
+pub struct CreateEventPayload {
+    // a null ref will occur on the initial create event
+    #[serde(rename = "ref")]
+    pub ref_field: Option<String>,
```
XAMPPRocky, line 39:
```suggestion
    pub r#ref: Option<String>,
```

tests/repos_events_test.rs
```diff
@@ -0,0 +1,143 @@
...
+            "documentation_url": "",
+            "errors": None::<Vec<serde_json::Value>>,
+            "message": message,
+        })))
+        .mount(&mock_server)
+        .await;
+}
+
+fn setup_octocrab(uri: &str) -> Octocrab {
+    OctocrabBuilder::new()
```
XAMPPRocky, line 55: Minor nit: You can just call `Octocrab::builder` which would remove the need to import `OctocrabBuilder`, but it's not super important.
```suggestion
    Octocrab::builder()
```

src/etag.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub etag: Option<Etag>,
+    /// The value identified by this etag.
+    ///
+    /// This can be `None` if we have already received the data which this etag identifies.
+    pub value: Option<T>,
+}
+
+/// An etag.
+#[derive(Debug, PartialEq)]
+pub struct Etag(String);
```
XAMPPRocky, line 22: I think it would be more useful to re-export the `hyperx::header::ETag` than creating our own type as we already depend on it for handling headers in pages. https://docs.rs/hyperx/1.2.1/hyperx/header/struct.ETag.html

src/models/events.rs
```diff
@@ -0,0 +1,294 @@
...
+    pub public: bool,
+    pub created_at: DateTime<Utc>,
+    pub payload: Option<EventPayload>,
+    pub org: Option<Org>,
+}
+
+/// The type of an event.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[non_exhaustive]
+pub enum EventType {
```
XAMPPRocky, line 26: Could you include a `#[serde(rename_all)]` attribute here, just so it's clear what their casing will be.

src/models/events.rs
```diff
@@ -0,0 +1,294 @@
...
+    data: serde_json::Value,
+) -> Result<Option<EventPayload>, serde_json::Error> {
+    let maybe_payload = match event_type {
+        EventType::PushEvent => serde_json::from_value::<PushEventPayload>(data)
+            .map(|payload| Some(EventPayload::PushEvent(payload)))?,
+        EventType::CreateEvent => serde_json::from_value::<CreateEventPayload>(data)
+            .map(|payload| Some(EventPayload::CreateEvent(payload)))?,
+        _ => serde_json::from_value::<serde_json::Value>(data)
+            .map(|payload| Some(EventPayload::UnknownEvent(payload)))?,
+    };
```
XAMPPRocky, line 119: I believe this match statement can be shortened to the following, though you should test it.
```suggestion
    let maybe_payload = Some(match event_type {
        EventType::PushEvent => serde_json::from_value(data)
            .map(EventPayload::PushEvent)?,
        EventType::CreateEvent => serde_json::from_value(data)
            .map(EventPayload::CreateEvent)?,
        _ => EventPayload::UnknownEvent(data),
    });
```

XAMPPRocky (changes requested):
Thank you for your PR! Overall this is great, and I'm favour of merging it (including the wiremock tests), I've left a couple of nit comments, and once those are resolved we can merge this in.

src/etag.rs
wayofthepie: Completely overlooked this, this is much nicer!

src/models/events.rs
wayofthepie: Yep much cleaner. Not sure how I overlooked the fact `data` is already a `serde_json::Value`, in the unknown case :smile:

wayofthepie:
Thanks for the feedback. I think I addressed all comments. Let me know if you want me to rebase into a single commit or anything before you merge.

Also, I had to remove `wiremock` and those tests in the end as it does not yet support 1.0. I can re-add once it does support it.

wayofthepie:
Scratch that on wiremock, it works fine on tokio 1.0, I have re-added those tests.

XAMPPRocky:
Thank you for your PR!

PR #64, opened 2021-01-08
Add rustls feature

Jake-Shadle:
Adds the `rustls` feature which in turn enables rustls for reqwest, and sets `reqwest/native-tls` as a default feature so that users will get the same behavior if they update octocrab.


XAMPPRocky:

PR #65, opened 2021-01-14
Add various enhancements

Jake-Shadle:
When using crate to create a Github action to automerge PRs, I ran into several missing pieces that I've added in this PR. These may seem a bit scattershot, so I can split out things in multiple PRs if you want me to.

1. Adds `PullRequestHandler::merge` which returns a `MergePullRequestsBuilder` that can be used to set the various options available when merging a PR.
1. Adds a `PullRequestHandler::list_reviews` to list all of the reviews that have been made on a PR.
1. Adds `RepoHandler::combined_status_for_ref` to retrieve all of the statuses for a particular reference
1. Adds various `params` and `models` for the above, as well as adding some missing fields (eg. `PullRequest::mergeable_state`)
1. Adds some models for data that are specific to [actions](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows), namely `PullRequestReviewAction` and `PullRequestAction`
1. Implemented a custom deserialize for `ReviewState`, as Github has a weird difference in that enumeration, as when making an API request, it is `SCREAMING_SNAKE_CASE` but when it's a webhook payload (which is what the action actually sees), it just `snake_case`

### DIFF
--- a/src/api/pulls.rs
+++ b/src/api/pulls.rs
@@ -203,7 +203,7 @@ impl<'octo> PullRequestHandler<'octo> {
     /// # Ok(())
     /// # }
     /// ```
-    pub async fn list_reviews(&self, pr: u64) -> crate::Result<Vec<crate::models::pulls::Review>> {
+    pub async fn list_reviews(&self, pr: u64) -> crate::Result<Page<crate::models::pulls::Review>> {
         let url = format!(
             "/repos/{owner}/{repo}/pulls/{pr}/reviews",
             owner = self.owner,
--- a/src/api/pulls.rs
+++ b/src/api/pulls.rs
@@ -2,6 +2,7 @@
 
 mod create;
 mod list;
+mod merge;
 
 use snafu::ResultExt;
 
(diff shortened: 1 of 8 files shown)


src/api/pulls.rs
```diff
@@ -194,6 +195,49 @@ impl<'octo> PullRequestHandler<'octo> {
...
     }
+
+    /// Lists all of the `Review`s associated with the pull request.
+    /// ```no_run
+    /// # async fn run() -> octocrab::Result<()> {
+    /// let reviews = octocrab::instance().pulls("owner", "repo").list_reviews(101).await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn list_reviews(&self, pr: u64) -> crate::Result<Vec<crate::models::pulls::Review>> {
```
XAMPPRocky, line 206: Shouldn't this return a `Page` and not just `Vec`? Presumably if it's PR with 100+ reviews this wouldn't give you all of them.

XAMPPRocky (commented):
Thank you for your PR! This all looks good to me, there's only one thing I have a question about.

src/api/pulls.rs
Jake-Shadle: Right, oops!

XAMPPRocky:
This has now been published in 0.8.7

PR #68, opened 2021-01-18
Support IssuesEvent

wayofthepie:
* Adds support for events of type `IssuesEvent`.
  * Factors out create and push events into their own modules under `payload`, the API stays the same, re-export from `payload`. But this makes it a bit cleaner, especially as the other event types get added.

Quick note, the events API is not documented very well on the official GitHub docs. For `IssuesEvent` events, [the docs](https://docs.github.com/en/developers/webhooks-and-events/github-event-types#issuesevent) specify 7 actions, however only 3 of those are available when calling the events API's, the rest (along with lots more) are only available via webhooks - see [this thread](https://github.com/igrigorik/gharchive.org/issues/183) for more info. 

I plan on adding support for webhooks once all event types are in octocrab anyway, so I've added all fields mentioned in the docs here, whether they are only available on webhooks or not. Thanks.

### DIFF
--- a/src/models/events.rs
+++ b/src/models/events.rs
@@ -1,6 +1,6 @@
 pub mod payload;
 
-use self::payload::{CreateEventPayload, EventPayload, PushEventPayload};
+use self::payload::{CreateEventPayload, EventPayload, IssuesEventPayload, PushEventPayload};
 use chrono::{DateTime, Utc};
 use reqwest::Url;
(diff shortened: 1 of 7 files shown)


XAMPPRocky:
Thank you for your PR!

XAMPPRocky:
Published as 0.8.8

PR #70, opened 2021-01-22
Support PullRequestEvent and PullRequestReviewCommentEvent

wayofthepie:
Also fixes the docs for push event, which had a bad ref back to the event type. Thanks.


src/models/events/payload.rs
```diff
@@ -24,6 +28,8 @@ pub enum EventPayload {
     CreateEvent(CreateEventPayload),
     IssuesEvent(IssuesEventPayload),
     IssueCommentEvent(IssueCommentEventPayload),
+    PullRequestEvent(Box<PullRequestEventPayload>),
+    PullRequestReviewCommentEvent(Box<PullRequestReviewCommentEventPayload>),
```
XAMPPRocky, line 32: Is there a particular reason these variants are boxed? I assume it has something to do with them containing `PullRequest`.

wayofthepie: Yep they are much larger than the rest of the variants. After boxing, a value of this enum is ~7KiB. Before boxing it's ~25KiB. Also, without boxing, both tests for deserializing - `should_deserialize_with_correct_payload` for both events - end up overflowing the stack when deserializing. 

Before boxing:
```
$ valgrind --tool=drd --show-stack-usage=yes target/debug/deps/octocrab-51c99cee680f94e2 \
    --test models::events::payload::pull_request::test::should_deserialize_with_correct_payload
...
thread 2 (which is a detached thread) finished and used 2112471 bytes out of 2093056 on its stack. Margin: -19415 bytes.
```
After boxing:
```
$ valgrind --tool=drd --show-stack-usage=yes target/debug/deps/octocrab-51c99cee680f94e2 \
    --test models::events::payload::pull_request::test::should_deserialize_with_correct_payload
...
test models::events::payload::pull_request::test::should_deserialize_with_correct_payload ... ok==309700== thread 2 (which is a detached thread) \
 finished and used 1337543 bytes out of 2093056 on its stack. Margin: 755513 bytes.
...
```
I'm not entirely sure why the stack usage drops almost 700KiB just by boxing those variants yet though! Seems like a lot, given that the payloads themselves are only ~25KiB.

XAMPPRocky:
Thank you for your PR!

XAMPPRocky:
Published in 0.8.10

PR #71, opened 2021-01-29
Support commit comment and delete events

wayofthepie:
These are small but odd as the actual events differ slightly from the docs:

  * `DeleteEvent` has a field `pusher_type` which is not documented. I have left this out for now (https://github.com/github/docs/issues/3192).
  * `CommitCommentEvent` has an action field documented, but it only appears in webhooks not when calling the API's. I have left this out for now too, it seems to only have a single value anyway, `created` (https://github.com/github/docs/issues/3190).

Thanks!

### DIFF
--- a/src/models/events.rs
+++ b/src/models/events.rs
@@ -1,9 +1,9 @@
 pub mod payload;
 
 use self::payload::{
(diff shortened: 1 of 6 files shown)


wayofthepie:
Woops :smile: didn't meant to close.

XAMPPRocky:
Thank you for your PR!

PR #72, opened 2021-02-02
Support member/gollum/fork events

wayofthepie:
These are also small. `MemberEvent` in reality has some slight differences from the docs (see https://github.com/github/docs/issues/3378). Thanks!


wayofthepie:
Ah just noticed it's failing on nightly. https://github.com/Alexhuszagh/rust-lexical/issues/55 seems to be the cause.

wayofthepie:
Oh, this PR slipped my mind, this is good now. You're probably aware, but the issue was nightly introduced a change that broke `lexical` (this broke octocrab in general, deps were wiremock -> nom -> lexical), nightly reverted it, and a few days ago `lexical` got a new release which fixes the issue. 

No rush with reviewing or anything, just commenting in case you wanted me to clarify the above!

XAMPPRocky:
Thank you for your PR!

PR #73, opened 2021-02-16
Support subdirectory-mounted API

udzura:
My company utilizes GHES and its v3 API is mounted under the url like `https://git.example.com/api/v3`

And this is usual way for GHES users, since GitHub itself shows as default below.

https://docs.github.com/en/enterprise-server@2.21/rest/overview/resources-in-the-rest-api#current-version

octocrab does not support subdirectory now(ignored if set). So it is very helpful for us that this patch is accepted.

Thank you!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -314,6 +314,17 @@ impl OctocrabBuilder {
 
         Ok(Octocrab {
             client,
+            path_prefix: match self.base_url.as_ref() {
+                Some(url) => {
+                    // In case of toplevel mount
+                    if url.path() == "/" {
+                        None
+                    } else {
+                        Some(url.path().to_string())
+                    }
+                }
+                None => None,
+            },
             base_url: self
                 .base_url
                 .unwrap_or_else(|| Url::parse(GITHUB_BASE_URL).unwrap()),
@@ -325,6 +336,7 @@ impl OctocrabBuilder {
 #[derive(Debug, Clone)]
 pub struct Octocrab {
     client: reqwest::Client,
+    path_prefix: Option<String>,
     pub base_url: Url,
 }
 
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! I'm in favour of adding it. However I'm a bit confused, is there a reason the following doesn't work as is?

```rust
crate::OctocrabBuilder::new()
                 .base_url("https://git.example.com/api/v3")
                 .unwrap()
                 .build()
                 .unwrap()
```

udzura:
Yes, we can specify a URL with subpath in `Builder::base_url()`. But subpaths are ignored when `Url::join` is called with paths starting `/...` e.g.

https://github.com/XAMPPRocky/octocrab/blob/master/src/api/issues/list.rs#L119

Minimum example:

```rust
let url = reqwest::Url::parse("https://git.example.com/api/v3").unwrap();
println!("start with / -> {}", url.join("/repos/foo/bar/issue").unwrap());
//=> start with / -> https://git.example.com/repos/foo/bar/issue
```

I don't find out this patch is the best way, but changing a large number of "absolute" paths to relative ones looks tough.

If you prefer to fix all of absolute paths, that should be also a good solution and I will follow.

udzura:
Wow, sorry for breaking compatibilities. This P/R occurs some degradation in case of below:

```rust
    let mut page = octocrab
        .issues(owner, repo)
        .list()
        .send()
        .await?;

    // page.next has submounted path e.g. `/api/v3/repos/foo/bar/issues?page=2` from API response
    // and octocrab::absolute_url() receive this path directry inside octocrab::get_page()
    // then will generate invalid API path
    while let Some(mut newpage) = octocrab.get_page(&page.next).await? {
        page = newpage;
    }
```

I'll close this PR and create another patch!

PR #74, opened 2021-02-17
Change absolutely specified API paths

udzura:
Related: #73 

Current octocrab does not support submounted API base url such as `"https://git.example.com/api/v3/"`.

This is because GitHub endpoints are coded to start with absolute `"/"` and passing `octocrab::absolute_url()` an absolute path will ignore `base_url`'s path info.

Note that only forms ending with `/`, e.g. `"https://git.example.com/api/v3/"` is accepted, not such like `"https://git.example.com/api/v3"`(This is url::Url's design). But it is OK for my use case.


XAMPPRocky:

PR #79, opened 2021-03-17
Fix: Release creation requires POST request, not GET

phil-opp:
See https://docs.github.com/en/rest/reference/repos#create-a-release

With the current GET request, the following error occurs:

```
Error: JSON Error in [0]: invalid type: map, expected a string representing an URL at line 1 column 1
```


XAMPPRocky:

PR #80, opened 2021-03-18
Support workflows/runs/jobs list API for github action

udzura:
* Listing workflows/runs/jobs for one repository.
* Added `workflows` namespace and `Octocrab::workflow()`, because actions API has lots of endpoints, so it seems better to be split by each purpose.
* Support yet other complicated JSON response forms like:

```javascript
{
  "total_count": 2,
  "jobs": [ // not "items" :( can be "workflows", "workflow_runs"
    {
      "id": 399444496,
      "run_id": 29679449,
      //...
    },
    {
      "id": 399444497,
      "run_id": 29679450,
      //...
    }
  ]
}
```

* APIs:
  * https://docs.github.com/en/rest/reference/actions#list-repository-workflows
  * https://docs.github.com/en/rest/reference/actions#list-workflow-runs
  * https://docs.github.com/en/rest/reference/actions#list-jobs-for-a-workflow-run

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -10,3 +10,4 @@ pub mod pulls;
 pub mod repos;
 pub mod search;
 pub mod teams;
+pub mod workflows;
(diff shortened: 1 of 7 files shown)


XAMPPRocky:
Thank you for your PR!

PR #82, opened 2021-03-21
List comments for repo and pr

terakilobyte:
Adds functionality to list comments in a PR. As part of this, a
Comment model was added, as well as a comment::Sort enum in params.

I'm unsure the best approach here. Using an Option for the pr number to switch between listing pull review comments for a repository vs listing for a specific pr doesn't feel wrong to me, but I did notice that you took a different approach to listing issue comments. I'm happy to change if requested.

### DIFF
--- a/src/api/pulls.rs
+++ b/src/api/pulls.rs
@@ -1,5 +1,6 @@
 //! The pull request API.
 
+mod comment;
 mod create;
 mod list;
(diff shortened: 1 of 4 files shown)


XAMPPRocky:
Thank you for your PR! This approach seems fine to me.

PR #83, opened 2021-04-20
Change committer/author type to User

zhaogang92:
There is already a User type defined, so I directly used that type (The original GitUser type actually matches the author/committer fields of the commit field, see https://docs.github.com/en/rest/reference/search#search-commits). The committer could be empty, so I make both author and committer to be an Option.

I checked all the other uses of GitUser and corresponding Github API, it seems only this place needs to be updated.


XAMPPRocky:

Issue #11, opened 2020-05-19
Refactor with builder procedural macro


XAMPPRocky:
Currently there's a lot of boilerplate in `octocrab` because of the builder pattern requires a lot of it to work properly, however there's no reason other than time and effort to replace the boilerplate with a procedural macro that generates the builder pattern code from the `struct` itself.

PR #94, opened 2021-06-10
Create custom macros to reduce boilerplate

dnaka91:
I was trying to add endpoints for repo webhooks and noticed the big amount of boilerplate for the used builders.
Saw that it is planned to reduce the boilerplate with custom macros as described in #11.

This is my attempt in creating said macros.

I created the PR as a draft for now and would like your input first on what you think about the current version before I go further with changes.
Personally I'm quite happy with the current solution, although there are a few more structs that have special cases that are not currently covered by the macros.

If you like the current version I'll maybe do a bit more cleanup in the macro implementation before I consider this ready for merging.


XAMPPRocky:
Thank you for your PR! However I've been working on a separate proc-macro crate to address this problem, that solves the problem in a slightly different way, that I think is a bit more novel that I want to use. Sorry for not communicating that in the issue, the crate is mostly finished but I've been procrastinating cleaning it up to properly support generics, and rewriting octocrab to use it as a test before making it public.

If you are interested I could make it public early without announcing, and we could collaborate on it and using it in Octocrab? This crate is more generally useful than this implementation. So you'd be able to use this in your own projects, it would be useful in any project where you have a long list optional parameters to a function.

dnaka91:
Oh that sounds interesting. Currently there are only `derive-builder` and `typed-builder` that work quite well but are lacking as well. So it would be great to have another contender with different approach to the builder pattern.

Building this macros was really fun though and I relearned a lot since I haven't done proc macros in a while. Also, at some point I thought too, "Just a few more things in here and this could be its own separate crate outside of octocrab" 😄.

On a site note, I was wondering what's the proper way of handling the lifetimes for the builders? I've seen different approaches throughout the code base and it doesn't seem to be 100% consistent (likely due to contributions from different people). Was just wondering what's the recommended way.

XAMPPRocky:
@dnaka91 Sorry for the delay. I've published the code https://github.com/XAMPPRocky/keywords The crate essenntially emulates keyword arguments in Rust, so you just define them in your function parameters, and then it generates a builder struct out of those parameters for people to call. Right now you can see in [`tests/basic.rs`](https://github.com/XAMPPRocky/keywords/blob/main/tests/basic.rs) I have two test cases commented out that I haven't gotten to work yet (namely working with generics, and working with a `impl` block. I would like to get those fixed before releasing.

```rust
#[keywords::r#fn]
fn foo(a: u32, b: u32, _: keywords! { c: u32, d: u32 }) {
    assert_eq!(a, c);
    assert_eq!(b, d);
}

foo(1, 2).c(1).d(2).call();
```

> On a site note, I was wondering what's the proper way of handling the lifetimes for the builders? I've seen different approaches throughout the code base and it doesn't seem to be 100% consistent (likely due to contributions from different people). Was just wondering what's the recommended way.

The correct answer is the most flexible, that's also valid, however by-reference makes builder's signifciantly more complicated, and one of the things I would hope to eventually address with `keywords`, is removing the boilerplate to take things by-reference.

XAMPPRocky:
Going to close this for inactivity. I still want this, if anyone wants to help finish `keywords` so that `octocrab` can use it. Feel free to open a new issue.

dnaka91:
Sorry, never really finished this. In the end I just implemented the few APIs by hand whenever I needed it.

PR #100, opened 2021-07-04
Add global events api

wayofthepie:
This adds support for the global events API. Thanks!

Note, this doesn't respect, or give a way to access, the `X-Poll-Interval` header which is documented [here](https://docs.github.com/en/rest/reference/activity#events). This is set to 60 seconds by default on the `events` API, which seems way too high given we can get ~80 new events per second on average from that API.

[This community post](https://github.community/t/fetching-events-from-public-event-api/120332) suggests that poll limit may not be enforced. I've yet to hit an issue with continual polling, as long as the rate limits are obeyed.

### DIFF
--- a/examples/poll_events.rs
+++ b/examples/poll_events.rs
@@ -1,0 +1,31 @@
+use octocrab::{etag::Etagged, models::events::Event, Page};
+use std::collections::VecDeque;
+
+const DELAY_MS: u64 = 500;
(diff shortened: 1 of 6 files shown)


XAMPPRocky:
Thank you for your PR! This looks good to me. I would like to expose `X-Poll-Interval`, but I think that can come later. Would you mind opening an issue for it?

PR #107, opened 2021-08-08
Change author/commiter type back to GitUser

hchockarprasad:
Changing committer/author type from `GitUser` to type `User` by PR #83  causes deserialization errors on few endpoints, as all fields that a `User` struct needs are not satisfied by the response data. One example is  [this](https://docs.github.com/en/rest/reference/repos#create-or-update-file-contents--code-samples) resource which provides data that cannot be converted to an `User` type.
So this PR roll-backs the changes done on #83  and set the committer/author type back to `GitUser`.


XAMPPRocky:

XAMPPRocky:
Published in 0.11.0

Issue #108, opened 2021-08-16
Panic when `documentation_url` is not in an error response


marier-nico:
Hello, and thanks for building this awesome crate!

I'm currently trying to use octocrab to list pull requests on repos, and I'm experiencing some issues with GitHub errors. The problem is that GitHub is sending back some HTTP 502 error codes, and when it does, the `documentation_url` field is not set in the request, which causes a panic.

<details>
  <summary><b>GitHub HTTP response</b></summary>

```
Response {
    url: Url {
        scheme: "https",
        cannot_be_a_base: false,
        username: "",
        password: None,
        host: Some(
            Domain(
                "api.github.com",
            ),
        ),
        port: None,
        path: "/repos/rust-lang/rust/pulls",
        query: Some(
            "state=closed&per_page=14",
        ),
        fragment: None,
    },
    status: 502,
    headers: {
        "server": "GitHub.com",
        "date": "Mon, 16 Aug 2021 19:34:05 GMT",
        "content-type": "application/json",
        "content-length": "32",
        "etag": "\"611ab88a-20\"",
        "vary": "Accept-Encoding, Accept, X-Requested-With",
        "x-github-request-id": "8583:5A9E:5B6593:1244875:611ABDA3",
    },
}
```
</details>

The panic is :
```
Error: HTTP Error: error decoding response body: missing field `documentation_url` at line 3 column 1

Caused by:
   0: error decoding response body: missing field `documentation_url` at line 3 column 1
   1: missing field `documentation_url` at line 3 column 1
```

<details>
  <summary><b>Full backtrace</b></summary>

  ```
Error: HTTP Error: error decoding response body: missing field `documentation_url` at line 3 column 1

Found at  0 <snafu::backtrace_shim::Backtrace as snafu::GenerateBacktrace>::generate::ha7667e3035162ea3
   /home/nmarier/.cargo/registry/src/github.com-1ecc6299db9ec823/snafu-0.6.10/src/backtrace_shim.rs:15
   <octocrab::error::Http as snafu::IntoError<octocrab::error::Error>>::into_error::h8bd1e90cb9c22d06
   /home/nmarier/.cargo/registry/src/github.com-1ecc6299db9ec823/octocrab-0.9.1/src/error.rs:5
 1 <core::result::Result<T,E> as snafu::ResultExt<T,E>>::context::{{closure}}::h807638accb6bcfd4
[...]

PR #109, opened 2021-08-17
Make documentation_url optional for `GitHubError`

marier-nico:
When the GitHub API returns HTTP 502 error codes, the `documentation_url` field is not present in the request body, which caused a panic when using octocrab. This PR makes the field optional, since it is not always returned.

Closes: #108


XAMPPRocky:

marier-nico:
Thanks for merging this in! :smile:

XAMPPRocky:
Published as 0.12.0

PR #119, opened 2021-09-17
Add support to get repositories content

crqra:
Adds support for the endpoint to [get repository content](https://docs.github.com/en/rest/reference/repos#get-repository-content).

**Changes:**

- Added `GetContentBuilder`
- Added `get_content` method on `RepoHandler`
- Added example to get a repository's content in `examples/get_content.rs`
- Changed type of `Content.download_url` from `String` to `Optional<String>` since not all content types have it (i.e. the directory content type has this field as `null` in the HTTP response)

**Notes:**

- I've kept the `ref` parameter as in the GitHub API for consitency: `r#ref` in Rust (LMK if you want me to change it to branch)

**Help Needed:**

As you can see, on the `GetContentBuilder.send` method I'm returning `models::repos::Content`. This works as expected for the cases where the content we're fetching is either a `file`, `symlink` or `submodule`. However if the specified `path` is for a directory, the deserialization fails (because the response is a list instead of a single object). I don't know how to make this work and would appreciate any help or guidelines on this, please.

I think it could be nice to have a `Content` enum that would be either a `File`, `Symlink`, `Submodule`, `Directory` or `Vec<Content>` (not sure if this is possible and how Serde handles this, although that would break the existing `Content` structure.


XAMPPRocky:
Thank you for your PR! 

> However if the specified path is for a directory, the deserialization fails (because the response is a list instead of a single object).

Is it a list or is a `Page`? As depending on which, it'd change my exact answer, but what you need to do, is implement `serde::Deserialize` for `Content` directly, and have the deserialize implementation check if there's one or more items in the response, and always store it in a `Vec`, then for the single item case it's a vec of one item.

If you want examples of how to have serde accept multiple types, I'd recommend looking at this example. https://serde.rs/string-or-struct.html

crqra:
> Is it a list or is a Page? 

It's a list. I went to double check and this endpoint isn't paginated. In the docs they say that in the case of directories the endpoint has an upper limit of 1000 entries. To get more files you'd have to use the [Git Tree API](). I went ahead and tried it out just to double check and in fact it only returns 1000 at most (that repo has almost 2000 in the root):

```bash
$ gh api repos/crqra/github-content-api-test/contents | jq '. | length'
1000
```

I'll go ahead and implement the `serde::Deserialize` for `Content` as per your suggestion.

Thanks for the useful information! 👍🏼

crqra:
I've ended up creating a `ContentItems` struct implementing the `FromResponse` trait (based on the `Page`). I've tried the implementing the `Deserialize` first on the `Content` with a custom `Deserializer`, but since it's a struct I couldn't return the `Vec` on it, which makes sense. Also tried to create a `Deserializer` for the `Vec<Content>` and learnt that we can only implement traits for arbitrary types if the traits are defined in the current crate.

Do you think this solution is acceptable or did I missed something with the attempt on the custom `Deserializer`?

src/models/repos.rs
```diff
@@ -64,13 +65,46 @@ pub struct Content {
...
+    }
+}
+
+#[async_trait::async_trait]
+impl crate::FromResponse for ContentItems {
+    async fn from_response(response: reqwest::Response) -> crate::Result<Self> {
+        let json: serde_json::Value = response.json().await.context(crate::error::Http)?;
+
+        if json.is_array() {
+            println!("x: {:?}", json);
```
XAMPPRocky, line 93:
```suggestion
```

XAMPPRocky:
Thank you for your PR!

PR #122, opened 2021-09-25
Add Support for Repository Creation from a Template Repository

okhuz:
For our automation we need a way to create a new service repository using our template services. This change enables a new repository creation from an existing repository if repository is a template. There are no is_template checks, github api is returning the message.


XAMPPRocky:

PR #128, opened 2021-10-19
Allow deserializing IDs as strings.

### LINKED ISSUE #123: Notifcation IDs fail to decode
I attempted to use your example code to get a list of the user notifications but it seems to be unable to decode the IDs. If I curl the API each ID is a string, whereas in the code the IDs are u64 and Serde doesn't seem to convert it.

judah:
Fixes #123. It turns out that the notifications API returns
NotificationIDs in JSON as strings, not integers (e.g.: `id1: "1234"`).
See for example:
https://docs.github.com/en/rest/reference/activity#list-notifications-for-the-authenticated-user--code-samples

This PR adds an explicit Deserializer for each ID type that allows
either raw integers or string-encoded integers. It fixes the behavior when I run
```
GITHUB_TOKEN=... cargo run --example notifications
```

(We could have done this just for NotificationID, but it seemed easier
to just do for every ID type, and perhaps will help future-proof this
library.)

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -5,7 +5,7 @@ use std::ops::{Deref, DerefMut};
 
 use chrono::{DateTime, Utc};
 use reqwest::Url;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, de, Deserializer};
 
 pub mod activity;
 pub mod events;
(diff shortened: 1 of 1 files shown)


XAMPPRocky:

PR #143, opened 2021-11-01
Make org in workflow run event optional

wayofthepie:
Just spotted this issue too. The org field in a workflow run event is optional. Said I'd keep it separate to the other PR. Also a breaking change.


XAMPPRocky:
Thank you for your PR!

PR #144, opened 2021-11-04
Add authenticated app request and models

mkroman:
This PR adds a `get()` method to the `AppsRequestHandler` that returns app details for the currently authenticated app.

Example:

```rust
let octocrab = Octocrab::builder().app(…, …).build()?;
let app = octocrab.apps().get().await?;

println!("app name: {}", app.name);
```

Currently marked as draft until #140 gets merged.


mkroman:
Maybe it's more appropriate if it's in `CurrentAuthHandler`?

XAMPPRocky:
Thank you for your PR! Having it in `current` makes sense to me, as then it could be `octocrab.current().app()`, rather than `octocrab.app().get()`, which I think is clearer.

mkroman:
I moved it to `CurrentAuthHandler` such that it is now `octocrab.current().app()`.

XAMPPRocky:

PR #145, opened 2021-11-06
Fix clippy lints

mkroman:
Running `cargo clippy` gave a bunch of warnings before, so I fixed them.

The reformatted code is just `rustfmt` doing its thing - I assume that's ok?


XAMPPRocky:
Thank you for your PR!

PR #147, opened 2021-11-11
Add support for listing all workflow runs in a repository

hayleyjames:
This adds supports for listing workflow runs by repository. The query parameters are the same whether you are listing by workflow id or repository, so it made sense to make the `ListRunsBuilder` reusable.

### DIFF
--- a/examples/list_all_workflow_runs.rs
+++ b/examples/list_all_workflow_runs.rs
@@ -1,0 +1,30 @@
+use octocrab::Octocrab;
+
+#[tokio::main]
+async fn main() -> octocrab::Result<()> {
+    let octocrab = Octocrab::builder().build()?;
+    let runs = octocrab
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR, however I also just merged #153, is this PR now not needed? Sorry I didn't get to yours first, but if it is functionally same, I'm concerned whether the returned object will be the same in all cases, GitHub has a habit of things being similar but different.

hayleyjames:
Unfortunately, #153 only lists the artifacts for a workflow run, while my PR lists all the workflow runs in a repository. As for the returned objects, both endpoints have the same return type and example in the [Github OpenAPI definition](https://raw.githubusercontent.com/github/rest-api-description/main/descriptions/api.github.com/api.github.com.json).

XAMPPRocky:
Thanks for clarifying, and thank you for your PR!

PR #150, opened 2021-11-15
Add webhook models

mkroman:
This adds models that are useful for creating webhooks.

The GitHub API has 2 types of webhooks - one for repositories and one for organizations. I'm not entirely sure if they differ in any way.

This should at least cover what's used for getting/creating an organization webhook.


XAMPPRocky:
Thank you for your PR!

PR #151, opened 2021-11-17
Add workflow dispatch

Jasper-Bekkers:
It looks like a `cargo fmt` run has also accidentally reformatted a bunch of unrelated code - leaving it in for now, but we can split it up later if needed.

Adds support for https://octokit.github.io/rest.js/v18#actions-create-workflow-dispatch

### DIFF
--- a/src/api/actions.rs
+++ b/src/api/actions.rs
@@ -291,6 +291,21 @@ impl<'octo> ActionsHandler<'octo> {
         self.crab.get(route, None::<&()>).await
     }
 
+    /// Lists artifacts for a workflow run. Anyone with read access to the 
+    /// repository can use this endpoint. If the repository is private you 
+    /// must use an access token with the `repo` scope. GitHub Apps must have 
+    /// the `actions:read` permission to use this endpoint.
+    ///
+    /// ```no_run
+    /// octo.actions()
+    ///    .create_workflow_dispatch("org", "repo", "workflow.yaml")
+    ///    // required
+    ///    .ref_field(&ref)
+    ///    // optional
+    ///    .inputs(serde_json::json!({"my-key": "my-value"}))
+    ///    .send()
+    ///    .await?;
+    /// ```
     pub fn create_workflow_dispatch(
         &self,
         owner: impl Into<String>,
--- a/src/api/actions.rs
+++ b/src/api/actions.rs
@@ -311,11 +311,13 @@ impl<'octo> ActionsHandler<'octo> {
         owner: impl Into<String>,
         repo: impl Into<String>,
         workflow_id: impl Into<String>,
+        r#ref: impl Into<String>,
     ) -> WorkflowDispatchBuilder<'_> {
         WorkflowDispatchBuilder::new(self.crab,
             repo.into(),
             owner.into(),
-            workflow_id.into()
+            workflow_id.into(),
+            r#ref.into()
         )
     }
 }
--- a/src/api/actions.rs
+++ b/src/api/actions.rs
@@ -13,14 +13,10 @@ pub struct WorkflowDispatchBuilder<'octo> {
 }
 
 impl<'octo> WorkflowDispatchBuilder<'octo> {
-    pub(crate) fn new(crab: &'octo Octocrab, owner: String, repo: String, workflow_id: String) -> Self {
-        Self { crab, owner, repo, workflow_id, data: Default::default() }
-    }
-
-    /// Required
-    pub fn ref_field(mut self, ref_attr: impl Into<String>) -> Self {
-        self.data.ref_field = ref_attr.into();
-        self
+    pub(crate) fn new(crab: &'octo Octocrab, owner: String, repo: String, workflow_id: String, r#ref: String) -> Self {
+        let mut this = Self { crab, owner, repo, workflow_id, data: Default::default() };
+        this.data.r#ref = r#ref;
+        this
     }
 
     /// Input keys and values configured in the workflow file. The maximum number of properties is 10.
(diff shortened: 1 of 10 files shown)


src/api/actions.rs
```diff
@@ -114,19 +159,16 @@ impl<'octo> ActionsHandler<'octo> {
...
+        let data_response =
+            if let Some(redirect) = response.headers().get(reqwest::header::LOCATION) {
+                let location = redirect.to_str().expect("Location URL not valid str");
+
+                self.crab._get(location, None::<&()>).await?
+            } else {
+                response
+            };
+
+        data_response.bytes().await.context(crate::error::Http)
```
Jasper-Bekkers, line 171: Unrelated change that's a bugfix for the cases where there's no redirect header, encountered this when calling this function.

src/api/actions.rs
```diff
@@ -1,8 +1,53 @@
...
+    }
+
+    /// Required
+    pub fn ref_field(mut self, ref_attr: impl Into<String>) -> Self {
+        self.data.ref_field = ref_attr.into();
+        self
+    }
+
+    /// Optional
+    pub fn inputs(mut self, inputs: serde_json::Value) -> Self {
```
XAMPPRocky, line 27: This should document that it panics, when it does, and include information on what the property does. You can mostly just copy from GitHub's own REST documentation. https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event

```suggestion
    /// Input keys and values configured in the workflow file. The maximum number of properties is 10.
    /// Any default properties configured in the workflow file will be used when inputs are omitted.
    ///
    /// # Panics
    /// If `inputs` is not `Value::Object`.
    pub fn inputs(mut self, inputs: serde_json::Value) -> Self {
```

src/api/actions.rs
```diff
@@ -1,8 +1,53 @@
...
+impl<'octo> WorkflowDispatchBuilder<'octo> {
+    pub(crate) fn new(crab: &'octo Octocrab, owner: String, repo: String, workflow_id: String) -> Self {
+        Self { crab, owner, repo, workflow_id, data: Default::default() }
+    }
+
+    /// Required
+    pub fn ref_field(mut self, ref_attr: impl Into<String>) -> Self {
+        self.data.ref_field = ref_attr.into();
+        self
+    }
```
XAMPPRocky, line 24: I'd remove `ref` as a separate method, as if it's a separate method, there's a chance you might miss it and only get that error at runtime, where as if we make it a required argument, this won't happen.

```suggestion
    pub(crate) fn new(crab: &'octo Octocrab, owner: String, repo: String, workflow_id: String, r#ref: String) -> Self {
        let mut this = Self { crab, owner, repo, workflow_id, data: Default::default() };
        this.data.r#ref = r#ref;
        this
    }
```

src/models/workflows.rs
```diff
@@ -93,3 +93,11 @@ pub struct Step {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub completed_at: Option<chrono::DateTime<chrono::Utc>>,
 }
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
+#[non_exhaustive]
+pub struct WorkflowDispatch {
+    #[serde(rename = "ref")]
+    pub ref_field: String,
```
XAMPPRocky, line 101: In general and in `octocrab`, I prefer to just use `r#` syntax rather than using `rename`. In my opinion it's easier and more to remember to add `r#`, rather than remember what this library called the field, even if it might less visually appealing. Especially when users are likely looking at the JSON or GitHub's API docs rather than octocrab's.
```suggestion
    pub r#ref: String,
```

src/api/actions.rs
```diff
@@ -240,6 +290,34 @@ impl<'octo> ActionsHandler<'octo> {
...
+    pub fn create_workflow_dispatch(
+        &self,
+        owner: impl Into<String>,
+        repo: impl Into<String>,
+        workflow_id: impl Into<String>,
+    ) -> WorkflowDispatchBuilder<'_> {
+        WorkflowDispatchBuilder::new(self.crab,
+            repo.into(),
+            owner.into(),
+            workflow_id.into()
```
XAMPPRocky, line 318:
```suggestion
        workflow_id: impl Into<String>,
        r#ref: impl Into<String>,
    ) -> WorkflowDispatchBuilder<'_> {
        WorkflowDispatchBuilder::new(self.crab,
            repo.into(),
            owner.into(),
            workflow_id.into(),
            r#ref.into()
```

XAMPPRocky (changes requested):
LGTM, just some minor changes around documentation and arguments.

XAMPPRocky:
@Jasper-Bekkers Seems like this just needs a small amount of cleanup in the tests before it can be merged.

komu:
Any chance of this getting merged? I need this functionality for my Slack-bot and I was going to make a PR myself, but then I noticed that there's already this one.

If necessary I could take a stab cleaning the code.

Jasper-Bekkers:
@komu I think all comments have been addressed, all this would need is a rebase I think, feel free to take a stab at that.

komu:
I made some changes:

* Cleaned up the changes made by `cargo fmt` (8 files unrelated to this PR are not included anymore)
* Fixed the text of `create_workflow_dispatch` doc comment
* Fixed the code of `create_workflow_dispatch` doc comment to be up to date and to compile
* Squashed the changes into a single commit

Other than that, I made no changes to the code. I opened a new PR #185 for the changes since I can't update this one.

PR #154, opened 2021-11-17
Add method to create organization webhook

mkroman:
It might be more convenient to be able to do:

```rust
let octocrab = octocrab::Octocrab::default();
octocrab.org("some-org").hooks().create(…).await?;
```

in the future, but for now I think it's overkill since there's 2 types of hooks which would lead to a lot of duplicate code.


XAMPPRocky:
Thank you for your PR!

PR #153, opened 2021-11-17
List all artifacts for a workflow run

Jasper-Bekkers:
Adds https://docs.github.com/en/rest/reference/actions#list-workflow-run-artifacts

PR #155, opened 2021-11-17
ETagged cleanup

Jasper-Bekkers:
This cleans up some of the duplicated code around creating ETagged pages, should be merged after #153 #151

Previously there was some boilerplate code around this, this should make it easier to make a few more requests ETagged with minimal code duplication.

Essentially this adds a `FromResponse` implementation for `Etagged<Page<T>>` and an `_get_with_etag` helper.


XAMPPRocky:
Thank you for your PR! Sorry for the late reply, it seems like this didn't pass CI, and the logs can't be recovered because it's been so long, so I'm going to close this for now. If you'd still like to see this merged feel free to make a new PR.

PR #157, opened 2021-11-21
Some minor fixes to CreateStatusBuilder

rocallahan:
Fixing the result of `send` is an incompatible API change unfortunately.


XAMPPRocky:
Thank you for your PR!

PR #165, opened 2021-11-25
fix list all issues/pulls example

Veeupup:
the latest docs of how to get all pages data has a bug, it is that it won't update the next page next URL after fetching a new page, this pr fix it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -45,7 +45,7 @@
 //!
 //! let octocrab = octocrab::instance();
 //! // Returns the first page of all issues.
-//! let page = octocrab.issues("octocrab", "repo")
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! Would you actually mind changing this example to use `Octocrab::all_pages` instead? It essentially does this loop for you.

Veeupup:
@XAMPPRocky resolved

Veeupup:
many thanks for your crate `octocrab`, I use it in [databend](https://github.com/datafuselabs/databend) to develop an interesting feature https://github.com/datafuselabs/databend/issues/2975, it simplifies my development.

XAMPPRocky:

PR #166, opened 2021-11-25
add issue_url in Comment model

Veeupup:
add issue_url in model `Comment`, we may need it to analyse which issue this comment related.


XAMPPRocky:
Thank you for your PR!

PR #181, opened 2022-01-11
Add request to list repositories starred by the authenticated user

kafji:
API doc: https://docs.github.com/en/rest/reference/activity#list-repositories-starred-by-the-authenticated-user

### DIFF
--- a/examples/list_repos_starred_by_authenticated_user.rs
+++ b/examples/list_repos_starred_by_authenticated_user.rs
@@ -1,0 +1,22 @@
+use octocrab::Octocrab;
+
+#[tokio::main]
+async fn main() -> octocrab::Result<()> {
+    let token = std::env::var("GITHUB_TOKEN").expect("GITHUB_TOKEN env variable is required");
+
+    let octocrab = Octocrab::builder().personal_token(token).build()?;
+
+    let repos = octocrab
+        .current()
+        .list_repos_starred_by_authenticated_user()
+        .sort("updated")
+        .per_page(100)
+        .send()
+        .await?;
+
+    for repo in repos {
+        println!("{}", repo.name);
+    }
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR! This mostly looks good to me, however it should be under `activity` directly and the method should be called `list_repos_starred_by_authenticated_user` to match octokit. https://octokit.github.io/rest.js/v18#activity-list-repos-starred-by-authenticated-user

kafji:
I've moved it to `activity::list_repos_starred_by_authenticated_user` and also renamed the request builder to `ListStarredReposBuilder`.

src/api/activity.rs
```diff
@@ -20,4 +20,83 @@ impl<'octo> ActivityHandler<'octo> {
...
     }
+
+    /// [List repositories starred by current authenticated user](https://docs.github.com/en/rest/reference/activity#list-repositories-starred-by-the-authenticated-user).
+    ///
+    /// ```no_run
+    /// # async fn run() -> octocrab::Result<()> {
+    /// octocrab::instance()
+    ///     .activity()
+    ///     .starring()
+    ///     .list_starred_repositories()
```
XAMPPRocky, line 31:
```suggestion
    ///     .list_repos_starred_by_authenticated_user()
```

kafji:
I just learned that I don't need to [modify the library for my needs](https://github.com/XAMPPRocky/octocrab#http-api).

Do you think we should keep the PR?

Since the user can add it on their end, I'm thinking we should close it.

XAMPPRocky:
Thank you for your PR! I think we should add it, having a typed API should be preferred if possible.

kafji:
Thank you for merging it.

PR #186, opened 2022-01-14
repos: get file content

MarcoIeni:
Add the ability to retrieve file content on a repository.

Example:
```rust
let mut content = octocrab::instance()
    .repos("XAMPPRocky", "octocrab")
    .get_content()
    .path("README.md")
    .r#ref("master")
    .send()
    .await
    .unwrap();
let contents = content.take_items();
let c = &contents[0];
let decoded_content = c.decoded_content().unwrap();
println!("{}", decoded_content);
```

PR #187, opened 2022-01-17
Get file contents when inlined by GitHub

gmambro:
On get_content, when the path is a file,  github response includes the actual file contents (unless it's too large) using the enconding and content fields.  Check https://docs.github.com/en/rest/reference/repos#get-repository-content for an example

This PR adds a (hopefully) quick way to get the data from the response.


gmambro:
closing because  duplicate of #186

PR #188, opened 2022-01-24
Add public key to repos api

spricet:
I didn't know if this technically belonged to the actions api but since `owner` and `repo` are already known here it felt like a natural place to put it.

### DIFF
--- a/src/api/repos.rs
+++ b/src/api/repos.rs
@@ -51,6 +51,23 @@ impl<'octo> RepoHandler<'octo> {
         self.crab.get(url, None::<&()>).await
     }
 
+    /// Get's a repository's public key.
+    /// ```no_run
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! I think it should go in `actions`, as that's where octokit places it. Generally as a rule of thumb, we try to follow Octokit's API since that is officially maintained. 🙂  https://octokit.github.io/rest.js/v18#actions-get-public-key

spricet:
Okay cool, thank you! I'll move it over. Are contributions like these useful? I am very impressed and happy with the way we can extend this api and have been leveraging that immensely. Would you prefer to keep the static api small and rely on extensions to cover these less common use cases?

XAMPPRocky:
Thank you for your PR!

PR #193, opened 2022-02-18
Bump jsonwebtoken to version 8.

khuey:
jsonwebtoken did drop its Debug impl on EncodingKey. The MSRV to use finish_non_exhaustive here is 1.53. Not entirely sure if that's acceptable to you.


XAMPPRocky:

PR #194, opened 2022-02-20
fix: unified to use raw identifier instead of rename

konojunya:
I found an implementation that uses [serde(rename)](https://serde.rs/field-attrs.html#rename) to use the `type` key, but the rest of the code basically uses the [raw identifiers](https://doc.rust-lang.org/rust-by-example/compatibility/raw_identifiers.html), so it needs to be fixed to align the implementation.


XAMPPRocky:

PR #195, opened 2022-02-21
Support for Updating gists via PATCH

mfelsche:
I did not add any tests, but this is according to the [github api reference](https://github.com/github/rest-api-description/blob/main/descriptions/api.github.com/api.github.com.yaml).

This has been manually tested against the github API and verified to be working.

I chose a fluent style for the builders. What might be controversial is the way updates to files are implemented, as the different update operations are too complex to sanely encode in a single builder method, i chose to switch with the `.file("filename")` method to a builder of an update operation on the file which exposes the same methods as the `UpdateGistBuilder` and delegates those to the this builder.

### DIFF
--- a/src/api/gists.rs
+++ b/src/api/gists.rs
@@ -18,6 +18,7 @@ impl<'octo> GistsHandler<'octo> {
     }
 
     /// Create a new gist.
+    ///
     /// ```no_run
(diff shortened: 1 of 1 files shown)


XAMPPRocky:

PR #201, opened 2022-04-18
Make gist description and gist file language nullable

LuaKT:
Hi, these fields cause a panic as they can be null. This makes them optional. Thanks!


XAMPPRocky:

PR #204, opened 2022-04-19
Make gist commit user and change status nullable

LuaKT:
Hi, it seems gist commit `change_status` values and `user` can be null, this fixes that.

Thanks


XAMPPRocky:
Thank you for your PR!

PR #206, opened 2022-04-20
Ensure nullable user on pull request comments can be parsed

### LINKED ISSUE #205: Listing pull requests comments response parsing fails for comment of deleted user
Listing comments from a pull request where there is at least one comment of a deleted user leads to a parsing error for the API response looking like this: `Error("invalid type: null, expected struct User", line: 0, column: 0)`.

Minimal example to reproduce with crate version 0.15.4:
```rust
#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
    octocrab::instance()
        .pulls("cucumber-rs", "cucumber")
        .list_comments(Some(31))
        .send()
        .await?;

    Ok(())
}
```

Looking at the JSON API response of this (https://api.github.com/repos/cucumber-rs/cucumber/pulls/31/comments) one can see the user is set to `null` for some comments. 

Changing the type of `user` from `User` to `Option<User>` in the following code (`models::pulls::Comment`) seems to solve this. Don't know whether it is intended like this.
https://github.com/XAMPPRocky/octocrab/blob/2c382ffdf2e9d0b0da1bf3b5f6565c8683c0de9e/src/models/pulls.rs#L240-L244

This could be related to #197 where there are also some fields set to null but not sure.

DerMagischeBier:
Ensures pull request comments can still be parsed properly in case the user of a comment is set to null. This seems to happen for comments of deleted users.

Fixes #205.

### DIFF
--- a/src/models/pulls.rs
+++ b/src/models/pulls.rs
@@ -239,7 +239,7 @@ pub struct Comment {
     pub original_commit_id: String,
     #[serde(default)]
     pub in_reply_to_id: Option<u64>,
-    pub user: User,
+    pub user: Option<User>,
     pub body: String,
     pub created_at: chrono::DateTime<chrono::Utc>,
     pub updated_at: chrono::DateTime<chrono::Utc>,


XAMPPRocky:
Thank you for your PR!

PR #208, opened 2022-04-23
Update dependencies to fix some warnings from cargo audit

autarch:
This updates the dep versions for `reqwest` and `tokio`.

This doesn't fix all the issues, because this crate still depends on chrono 0.4.19, which as a potential segfault. But there's no newer version of chrono yet. See https://github.com/chronotope/chrono/issues/674 for discussion of the 0.4.20 release.


XAMPPRocky:
Thank you for your PR!

PR #210, opened 2022-04-29
Add a function that exposes the installation token

mbirtwell:
You can use the installation token to access git repos using a regular git
client, so it's useful to expose this beyong the octocrab client once it
has been generated.

I've added a `installation_and_token` function which calls `installation` to
create an `Octocrab` instance for the installation and also forces the cache
of the installation token to be populated then returns both. This means that
the user has to get the token at the same time as they're getting the
installation client. But also that you can't try and get an installation
token from a client that isn't configured for authorising as an installation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -467,6 +467,21 @@ impl Octocrab {
             },
         }
     }
+
+    /// Similar to `installation`, but also eagerly caches the installation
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR!

PR #219, opened 2022-06-04
Add serialize deserialize for ContentItems

frolovdev:
According to best practices, it's a good idea to make public-facing data structures serializable and deserializeble 

https://rust-lang.github.io/api-guidelines/interoperability.html#data-structures-implement-serdes-serialize-deserialize-c-serde

### DIFF
--- a/src/models/repos.rs
+++ b/src/models/repos.rs
@@ -74,7 +74,7 @@ pub struct Content {
     pub license: Option<License>,
 }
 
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! Sorry for the late reply but it seems like there are conflicting implementations with adding `Deserialize` and the `FromResponse` trait. Would you be able to fix that and then I can merge this?

PR #222, opened 2022-06-25
Add into_stream method to Page

### LINKED ISSUE #214: `impl futures::Stream for Page`
Currently paging can be a bit awkward, it would be nice if we could implement a stream API for pages.

mbirtwell:
More or less the easy version of what we discussed in #214.

Except that I've hidden this behind the stream feature. It's roughly orthogonal to the existing code that's behind the stream feature. But it requires exactly the same additional dependencies as that does. 

Closes #214

### DIFF
--- a/src/page.rs
+++ b/src/page.rs
@@ -1,16 +1,18 @@
-#[cfg(feature = "stream")]
-use futures_core::Stream;
-#[cfg(feature = "stream")]
-use futures_util::stream::try_unfold;
 use std::slice::Iter;
 
-#[cfg(feature = "stream")]
-use crate::Octocrab;
 use hyperx::header::TypedHeaders;
-#[cfg(feature = "stream")]
-use serde::de::DeserializeOwned;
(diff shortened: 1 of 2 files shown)


src/page.rs
```diff
@@ -29,6 +37,13 @@ pub struct Page<T> {
     pub last: Option<Url>,
 }
 
+#[cfg(feature = "stream")]
```
XAMPPRocky, line 40: This doesn’t need a cfg because it doesn’t use any stream specific stuff

XAMPPRocky (commented):
Mostly LGTM, just some small tweaks for clarity.

src/page.rs
mbirtwell: But if we don't put the cfg on it we'll get a dead code warning, which I wanted to avoid.

XAMPPRocky: Ah yeah, then it's fine.

XAMPPRocky:

PR #223, opened 2022-06-28
Fix absolute API path for workflow API to support path prefixed base_url

s4ichi:
related fix: https://github.com/XAMPPRocky/octocrab/pull/74

The current implementation of workflow runs API does not consider `base_url` that has a path prefix.

https://github.com/XAMPPRocky/octocrab/blob/807b501c0e3e67bc52b98d9a69f594e78ad982f4/src/api/workflows.rs#L253-L270

This PR modifies that path fixed in the same way as the related PR, so that it can use a path prefix.


XAMPPRocky:

PR #231, opened 2022-08-02
Make all UpdateGistFile fields optional

modhanami:
GitHub REST API allows omitting unchanged fields of a gist file but doesn't allow them to be explicitly null.

This PR fixes that by skipping serializing these fields when they are null.


XAMPPRocky:

PR #234, opened 2022-08-11
Fix not clearing cached token on 401

qbx2:
reqwest::send does not return Err even if server returned 401. We should explicitly call Response::error_for_status to convert 401 to Err.


XAMPPRocky:

PR #239, opened 2022-08-16
Fix 401 handler

### LINKED ISSUE #236: Is it still possible to retrieve the error reason when GitHub returns a 4xx status code?
GitHub's API returns an error message even when it returns a 4xx status code.  For example, doing `https://api.github.com/search/code?q=test` without authentication will result in a `Validation Failed` message.  Other more useful statuses include the `secondary rate limit` status, if too many requests are made within a short period of time.

Previously, it was possible to retrieve the response JSON of a GitHub API error by doing `let resp = octo._get(url, parameters).await?;`, and then doing `octocrab::map_github_error(resp).await` to get the GitHub error object.  By doing that, we could put together code that retries when we are rate limited, for example:
```rust
fn is_rate_limited(err: &octocrab::Error) -> bool {
    match err {
        octocrab::Error::GitHub {
            source,
            backtrace: _,
        } => source.message.contains("secondary rate limit"),
        _ => false,
    }
}

async fn get_and_retry_if_rate_limited<T: FromResponse, P: serde::Serialize + ?Sized>(
    octo: &Octocrab,
    url: &str,
    parameters: Option<&P>,
) -> std::result::Result<T, Box<dyn std::error::Error>> {
    loop {
        let resp = octo._get(url, parameters).await?;
        let ret = match match octocrab::map_github_error(resp).await {
            Err(err) => Err(err),
            Ok(resp) => T::from_response(resp).await,
        } {
            Ok(ret) => ret,
            Err(err) => {
                if is_rate_limited(&err) {
                    let retry_secs = 30;
                    println!(
                        "Being rate limited, retrying after {} seconds ...",
                        retry_secs
                    );
                    thread::sleep(time::Duration::from_secs(retry_secs));
                    continue;
                } else {
                    Err(err)?
                }
            }
        };
        return Ok(ret);
    }
}
```

The code above does not work since https://github.com/XAMPPRocky/octocrab/pull/234 was merged, and there does not seem to be any other way to do this.  Is this really intended?

qbx2:
Close #236

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -825,14 +825,15 @@ impl Octocrab {
                 )
                 .bearer_auth(app.generate_bearer_token()?)
                 .send()
-                .await
-                .and_then(|r| r.error_for_status());
-            if let Err(ref e) = result {
-                if let Some(StatusCode::UNAUTHORIZED) = e.status() {
-                    if retries < MAX_RETRIES {
-                        retries += 1;
-                        continue;
-                    }
+                .await;
+            let status = match &result {
+                Ok(v) => Some(v.status()),
+                Err(e) => e.status(),
+            };
+            if let Some(StatusCode::UNAUTHORIZED) = status {
+                if retries < MAX_RETRIES {
+                    retries += 1;
+                    continue;
                 }
             }
             let response = result.context(error::HttpSnafu)?;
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR!

PR #240, opened 2022-08-19
disable default-features on chrono to avoid triggering CVE-2020-26235 on audit

HayleyDeckers:
The Chrono library depends on `time v0.1.43` when build with the default features. This will trigger a warning of CVE-2020-26235 when a project that depends on Octocrab uses `cargo-deny` or `cargo-audit`.

This PR disables the default-features and instead pulls in the "clock" feature for the `::now()` function, avoiding the warning.


XAMPPRocky:

PR #243, opened 2022-09-15
Add `list_branches` method to `repos` module

chachako:
First of all, this library is great! But it seems that a critical [API](https://docs.github.com/en/rest/branches/branches#list-branches) is missing, so I simply added it and attached a small test, merge or close at will :)

### DIFF
--- a/src/api/repos/branches.rs
+++ b/src/api/repos/branches.rs
@@ -52,24 +52,3 @@ impl<'octo, 'r> ListBranchesBuilder<'octo, 'r> {
         self.handler.crab.get(url, Some(&self)).await
     }
 }
-
-#[cfg(test)]
-#[tokio::test]
-async fn test() {
-    let octocrab = crate::instance();
-    let branches = octocrab
-        .repos("rust-lang", "rust")
-        .list_branches()
-        .send()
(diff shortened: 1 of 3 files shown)


src/api/repos/branches.rs
```diff
@@ -0,0 +1,75 @@
...
+#[cfg(test)]
+#[tokio::test]
+async fn test() {
+    let octocrab = crate::instance();
+    let branches = octocrab
+        .repos("rust-lang", "rust")
+        .list_branches()
+        .send()
+        .await
+        .unwrap();
```
XAMPPRocky, line 65: Don't have the test actually make a connection to GitHub, as otherwise we run into rate limit issues like what's happened in CI. You can either remove the test or use a mock test like can be seen elsewhere.

XAMPPRocky (commented):
Thank you for your PR! This looks good, we just need to adjust the test to actually use the GitHub API to pass CI.

src/api/repos/branches.rs
chachako: I deleted the test now

XAMPPRocky:

PR #245, opened 2022-09-21
Add encoding field to content

suryapandian:
As per [the doc in github](https://docs.github.com/en/rest/repos/contents#get-repository-content) for files that are large i.e) 1-100MB "the content field will be an empty string and the encoding field will be "none"."

When encoding is none, the user can then use it to download the content via download_url.
Having this field, lets the octocrab sdk be in tune with the API doc of github.


XAMPPRocky:

PR #248, opened 2022-10-03
Add OAuth + device  flow

### LINKED ISSUE #247: Support device authentication flow
The project I'm working on requires a [`device flow`](https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#device-flow) login. I have a solution that works with octocrab to handle this. Would `octocrab` be interested in supporting this authentication workflow directly? If so I can submit a PR.

iwahbe:
Add support for Github's device flow. The document describing the flow is at https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#device-flow.

Fixes #247

### DIFF
--- a/examples/device_flow.rs
+++ b/examples/device_flow.rs
@@ -1,0 +1,40 @@
+use std::time::Duration;
+const CLIENT_ID: &'static str = "********************";
+
+#[tokio::main]
+async fn main() -> octocrab::Result<()> {
+    let codes = octocrab::auth::authenticate_as_device(
+        secrecy::Secret::from(String::from(CLIENT_ID)),
+        ["public_repo", "read:org"],
+    )
+    .await?;
+    println!(
+        "Go to {} and enter code {}",
+        codes.verification_uri, codes.user_code
+    );
+    let mut interval = Duration::from_secs(codes.interval);
+    let mut clock = tokio::time::interval(interval);
+    let auth = loop {
+        clock.tick().await;
+        match codes.poll_once().await? {
+            Ok(auth) => break auth,
+            Err(cont) => match cont {
+                octocrab::auth::Continue::SlowDown => {
+                    // We were request to slow down. We add five seconds to the polling
+                    // duration.
+                    interval += Duration::from_secs(5);
+                    clock = tokio::time::interval(interval);
+                    // The first tick happens instantly, so we tick that off immediately.
+                    clock.tick().await;
+                }
+                octocrab::auth::Continue::AuthorizationPending => {
+                    // The user has not clicked authorize yet, but nothing has gone wrong.
+                    // We keep polling.
+                }
+            },
+        }
+    };
+
+    println!("Authorization succeeded with access to {:?}", auth.scope);
+    return Ok(());
+}
--- a/examples/device_flow.rs
+++ b/examples/device_flow.rs
@@ -1,13 +1,18 @@
+use either::Either;
+use reqwest::header::ACCEPT;
 use std::time::Duration;
-const CLIENT_ID: &'static str = "********************";
 
 #[tokio::main]
 async fn main() -> octocrab::Result<()> {
-    let codes = octocrab::auth::authenticate_as_device(
-        secrecy::Secret::from(String::from(CLIENT_ID)),
-        ["public_repo", "read:org"],
-    )
-    .await?;
+    let client_id = secrecy::Secret::from(std::env::var("GITHUB_CLIENT_ID").unwrap());
+    let crab = octocrab::Octocrab::builder()
+        .base_url("https://github.com")?
+        .add_header(ACCEPT, "application/json".to_string())
+        .build()?;
+
+    let codes = crab
+        .authenticate_as_device(&client_id, ["public_repo", "read:org"])
+        .await?;
     println!(
         "Go to {} and enter code {}",
         codes.verification_uri, codes.user_code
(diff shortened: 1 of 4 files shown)


src/auth.rs
```diff
@@ -87,3 +90,166 @@ impl AppAuth {
...
+            token_type: value.token_type,
+            scope: value.scope.split(",").map(ToString::to_string).collect(),
+        }
+    }
+}
+
+/// Authenticate with Github's device flow. This starts the process to obtain a new `OAuth`.
+///
+/// See https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#device-flow for details.
+pub async fn authenticate_as_device<I, S>(client_id: SecretString, scope: I) -> Result<DeviceCodes>
```
XAMPPRocky, line 124: This should not be a free function, it should be an inherent method on `Octocrab`, otherwise this code won’t work with enterprise installations which use different base URLs.

examples/device_flow.rs
```diff
@@ -0,0 +1,40 @@
+use std::time::Duration;
+const CLIENT_ID: &'static str = "********************";
```
XAMPPRocky, line 2: This should be retrieved through `std::env::var` using the variable `GITHUB_CLIENT_ID` rather than using a const.

src/auth.rs
```diff
@@ -87,3 +90,166 @@ impl AppAuth {
...
+    /// The minimum number of seconds that must pass before you can make a new access
+    /// token request (POST https://github.com/login/oauth/access_token) to complete the
+    /// device authorization. For example, if the interval is 5, then you cannot make a
+    /// new request until 5 seconds pass. If you make more than one request over 5
+    /// seconds, then you will hit the rate limit and receive a slow_down error.
+    pub interval: u64,
+
+    // Used to poll for the device codes.
+    #[serde(skip)]
+    crab: crate::Octocrab,
```
XAMPPRocky, line 179: This shouldn’t be included with the rest of the struct.

src/auth.rs
```diff
@@ -87,3 +90,166 @@ impl AppAuth {
...
+
+    // Used for polling the device. This should be filled in as soon as possible.
+    #[serde(skip)]
+    client_id: Option<SecretString>,
```
XAMPPRocky, line 183: This should not be stored, and instead passed as an argument.

src/auth.rs
```diff
@@ -87,3 +90,166 @@ impl AppAuth {
...
+}
+
+impl DeviceCodes {
+    /// Poll Github to see if authentication codes are available.
+    ///
+    /// See `https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#response-parameters` for details.
+    pub async fn poll_once(&self) -> Result<Result<OAuth, Continue>> {
```
XAMPPRocky, line 190: This should be implemented on `Octocrab` and have device codes and client_id as arguments. This should also use `either::Either` for the inner type rather than result.

XAMPPRocky (commented):
Thank you for your PR! This mostly looks good to me, there’s mostly some ergonomic improvements needed to make it consistent with the rest of the API.

src/auth.rs
iwahbe: I didn't want to add any dependencies on my PR, but I agree that this is better. I'll make that change.

src/auth.rs
iwahbe: I wanted a way to do this without requiring the user to reconfigure a second builder. I now see [Octocrab::builder](https://docs.rs/octocrab/latest/octocrab/struct.Octocrab.html#implementations). I'll make this change.

src/auth.rs
```diff
@@ -87,3 +90,171 @@ impl AppAuth {
...
+    /// ```no_run
+    /// # async fn run() -> octocrab::Result<()> {
+    /// # use reqwest::header::ACCEPT;
+    /// let crab = octocrab::Octocrab::builder()
+    /// .base_url("https://github.com")?
+    /// .add_header(ACCEPT, "application/json".to_string())
+    /// .build()?;
+    /// # Ok(())
+    /// # }
+    /// ```
```
iwahbe, line 138: I feel like the default should just work here. Right now this always fails:
```rust
let crab = Octocrab::default();
crab.authenticate_as_device(client_id, scopes).await?;
```

I wish there was a way to make the default just work. One idea is introspecting `self` and correcting missing accept headers and the default base url (which will never work). We wouldn't do anything if the user overwrite the base url. What do you think?

src/auth.rs
```diff
@@ -87,3 +90,172 @@ impl AppAuth {
...
+
+impl DeviceCodes {
+    /// Poll Github to see if authentication codes are available.
+    ///
+    /// See `https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#response-parameters` for details.
+    pub async fn poll_once(
+        &self,
+        crab: &crate::Octocrab,
+        client_id: &SecretString,
+    ) -> Result<Either<OAuth, Continue>> {
```
iwahbe, line 201: I personally like exposing `poll_once` only after `authenticate_as_device` has been called. The other option is exposing the function as 
```rust
Octocrab::poll_for_device_authentication(&self, codes: &DeviceCodes, client_id: &SecretString)
```

I'm not sure which one would be more idiomatic in the crate.

iwahbe (commented):
Sorry for being slow to respond. I made fixes responding to your comments. I commented on two areas that I thought need more consideration.

src/lib.rs
```diff
@@ -336,6 +336,15 @@ impl OctocrabBuilder {
                 AuthState::None
             }
             Auth::App(app_auth) => AuthState::App(app_auth),
+            Auth::OAuth(device) => {
```
rsdy, line 339: I'm currently testing the API for this PR for a project, and I can't find an obvious way to actually set this value.

[`OctocrabBuilder`](https://docs.rs/octocrab/latest/octocrab/struct.OctocrabBuilder.html) has `app`, `basic_auth`, and `personal_token` methods that set the other values for `auth`, but this PR doesn't seem to provide a way to set `auth` on the builder instance.

rsdy:
Is there a way I could help get this PR through the finish line? Would be happy to contribute.

PR #253, opened 2022-10-14
`Issue.state` has type `IssueState`, not `String`

coriolinus:
Strong types are better types.


XAMPPRocky:
Thank you for your PR!

PR #256, opened 2022-10-15
Adding invitations.

manchicken:


manchicken:
Ok, I think I got the tests in right this time as well. I'm feeling a bit more confident with this one, but I learned not to copy-and-paste the full JSON payload from GitHub's documentation since sometimes their documentation is straight-up wrong (`failed_at` doesn't send an empty string, it sends `null` when empty).

XAMPPRocky:
Thank you for your PR!

PR #260, opened 2022-10-30
Graduate issues lock API

MarcoFalke:
The API graduated, see https://developer.github.com/changes/1/#sailor-v-preview-graduates

Also, solve "See also" URL redirects. Not sure if they are still needed, given that the other methods don't have them?


XAMPPRocky:
Thank you for your PR!

PR #261, opened 2022-10-30
Make Issue.number type u64

MarcoFalke:
This is needed because the `IssueHandler` assumes `u64` in all methods, see for example https://docs.rs/octocrab/latest/octocrab/issues/struct.IssueHandler.html#method.update

This should avoid having to type `number.try_into().unwrap()` when using the member variable.


MarcoFalke:
Just to mention the obvious: This is a breaking change, so some user code may need to be adjusted after this.

XAMPPRocky:

PR #267, opened 2022-11-11
Logging option for debugging (`connection_verbose`)

dahlia:
This patch adds `OctocrabBuilder::connection_verbose()`, which enables (or disables) underlying [`reqwest::Client`'s `connection_verbose` option][1].  Turning on this makes all requests and responses are logged, which is useful for debugging.

[1]: https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.connection_verbose


dahlia:
Resolved conflicts and rebased on the latest master.

PR #272, opened 2022-11-23
Implement reactions for issues

aureleoules:
This PR allows to:
- list reactions of an issue / issue comment
- create a reaction on an issue / issue comment
- delete a reaction on an issue / issue comment

### DIFF
--- a/src/api/issues.rs
+++ b/src/api/issues.rs
@@ -775,3 +775,346 @@ impl<'octo, 'r> ListIssueCommentsBuilder<'octo, 'r> {
         self.handler.crab.get(route, Some(&self)).await
     }
 }
+
+impl<'octo> IssueHandler<'octo> {
+    /// Lists reactions for an issue.
+    /// ```no_run
+    /// # async fn run() -> octocrab::Result<()> {
+    /// let reactions = octocrab::instance()
+    ///     .issues("owner", "repo")
+    ///     .list_reactions(1)
+    ///     .per_page(100)
+    ///     .page(2u32)
+    ///     .send()
+    ///     .await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn list_reactions(&self, issue_number: u64) -> ListReactionsBuilder<'_, '_> {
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR!

src/api/issues.rs
```diff
@@ -775,3 +775,346 @@ impl<'octo, 'r> ListIssueCommentsBuilder<'octo, 'r> {
...
+
+        self.handler.crab.get(route, Some(&self)).await
+    }
+}
+
+#[derive(serde::Serialize)]
+pub struct ListCommentReactionsBuilder<'octo, 'r> {
+    #[serde(skip)]
+    handler: &'r IssueHandler<'octo>,
+    comment_id: u64,
```
MarcoFalke, line 865: Shouldn't the type be `CommentId`? Otherwise a user might have to use extra code to change the type in their code?

aureleoules: I followed the way the current code does it, example https://github.com/XAMPPRocky/octocrab/blob/master/src/api/issues.rs#L658

XAMPPRocky: Followup PR always welcome 🙂

src/models/reactions.rs
```diff
@@ -0,0 +1,31 @@
...
+    #[serde(rename = "eyes")]
+    Eyes,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[non_exhaustive]
+pub struct Reaction {
+    pub id: ReactionId,
+    pub node_id: String,
+    pub user: User,
```
MarcoFalke, line 28: This might be optional, no? According to a schema https://docs.github.com/en/rest/reactions#create-reaction-for-an-issue:

```
"user": {
      "anyOf": [
        {
          "type": "null"
        },
        {
          "title": "Simple User",
...

src/api/issues.rs
MarcoFalke: Thanks, done in https://github.com/XAMPPRocky/octocrab/pull/274

PR #273, opened 2022-11-23
Some improvements to README.md

SkyfallWasTaken:
Hi!

This PR changes some stuff in the readme:

- Update version `0.16` to the latest, `0.17`
- Make `docs.rs` links point to version 0.17, in case a new version of Octocrab is released that contains breaking changes
- Change the repo used in the "Getting a Pull Request" section to the Octocrab repo, since you use it in the "HTTP API" section

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 [![Help Wanted](https://img.shields.io/github/issues/XAMPPRocky/octocrab/help%20wanted?color=green)](https://github.com/XAMPPRocky/octocrab/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22)
 [![Lines Of Code](https://tokei.rs/b1/github/XAMPPRocky/octocrab?category=code)](https://github.com/XAMPPRocky/octocrab)
 [![Documentation](https://docs.rs/octocrab/badge.svg)](https://docs.rs/octocrab/)
+[![Crates.io](https://img.shields.io/crates/v/octocrab?logo=rust)](https://crates.io/crates/octocrab/)
 
 Octocrab is a third party GitHub API client, allowing you to easily build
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,7 +13,7 @@ semantic API, and a lower level HTTP API for extending behaviour.
 
 #### Cargo.toml
 ```toml
-octocrab = "0.16"
+octocrab = "0.17"
```
XAMPPRocky, line 16: I think I'd rather point to the new cargo command which is evergreen, and keep the links below to latest.

```suggestion
cargo add octocrab
```

XAMPPRocky (changes requested):
Thank you for your PR! However I'd think we should make a slight change to reduce the number of times this is updated.

README.md
lineville: Or maybe just including this badge which will always show the current published cargo version?

crates.io

SkyfallWasTaken:
@XAMPPRocky @lineville 

Just added both of your changes, please take a look.

XAMPPRocky:
@SkyfallWasTaken Thank you for doing that, you also just need to remove the changes to the docs.rs URLs, and then it can be merged.

SkyfallWasTaken:
@XAMPPRocky Just removed them, PTAL

One thing I did decide to do is point out that the module list is valid as of version 0.17 (in case the README isn't updated when the version is changed):

```markdown
The semantic API provides strong typing around GitHub's API, a set of
[`models`] that maps to GitHub's types, and [`auth`] functions that are useful
for GitHub apps.
Currently, the following modules are available as of version `0.17`.
```

Is that OK, or should I remove that too?

XAMPPRocky:

PR #277, opened 2022-11-27
Add 'request_reviews' function

aureleoules:
Adds a command to request reviews from reviewers. See https://docs.github.com/en/rest/pulls/review-requests#request-reviewers-for-a-pull-request.

I had to add 'Open' to the `ReviewState` enum, otherwise, there would be a deserialization error after a review request.

### DIFF
--- a/src/api/pulls.rs
+++ b/src/api/pulls.rs
@@ -239,6 +239,36 @@ impl<'octo> PullRequestHandler<'octo> {
         self.http_get(url, None::<&()>).await
     }
 
+    /// Request a review from users or teams.
+    /// ```no_run
+    /// # async fn run() -> octocrab::Result<()> {
+    /// use octocrab::params;
+    /// let review = octocrab::instance().pulls("owner", "repo")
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR!

src/api/pulls.rs
```diff
@@ -239,6 +239,36 @@ impl<'octo> PullRequestHandler<'octo> {
...
+    /// let review = octocrab::instance().pulls("owner", "repo")
+    ///    .request_reviews(101, vec!["user1".to_string(), "user2".to_string()], vec!["team1".to_string(), "team2".to_string()])
+    ///  .await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn request_reviews(
+        &self,
+        pr: u64,
+        reviewers: Vec<String>,
```
MarcoFalke, line 255: Why not use `impl Into Vec<String>`? This should allow to drop the `vec!` above.

MarcoFalke (commented):
thanks, lgtm

PR #278, opened 2022-11-28
Adding debugging to response

manchicken:
This PR gives you a quick `DEBUG_OCTOCRAB` environment variable, which when equal to `"1"`, will cause errors in responses to print out the HTTP response body.


manchicken:
I think it's still relevant; hadn't noticed the conflicts arise. I haven't gotten much feedback on the PR, but I do think it still adds value. 

I'll look into the conflicts.

XAMPPRocky:
Thank you for your PR! Sorry for the delay in reviewing it. I'm all for the making the response more debuggable, but I'm not sure about the implementation, using `println` and `env` are simple but they don't work well with other systems or environments, and I'd like to eventually make this crate `#[no_std]` so that it works without any I/O dependencies.

With the tower layer stuff merged now, is there anything that this would provide over using tracing and setting `RUST_LOG=octocrab=debug`? That fits in a more standard way of how crates provide debug information.

XAMPPRocky:
I'm going to close this for inactivity, and it's no longer relevant. You can now customise your HTTP client if you specifically want this. In octocrab we should be focused on the GitHub aspect of debugging.

PR #279, opened 2022-12-01
request_reviews fixups

MarcoFalke:
Address my review comments from https://github.com/XAMPPRocky/octocrab/pull/277#pullrequestreview-1195552584

I also tried to get it to accept a vec/slice/array of `&str`, but didn't see an easy way to achieve this.


XAMPPRocky:
Thank you for your PR!

PR #280, opened 2022-12-06
crash when using pagination on `installation/repositories`

gabm:
When doing

```rust
            let first_repo_page: Page<Repository> = octocrab
                .get("/installation/repositories", None::<&()>)
                .await?;

            let repo_pages = octocrab.all_pages(first_repo_page).await?;
```

octocrab crashes here: https://github.com/XAMPPRocky/octocrab/blob/99872f252184b3a509b05a4e118a5245b51cce79/src/page.rs#L185-L188

Shortterm solution is to add the repositories attribute to the expected attributes list. In any case, the whole list feels shaky, this will likely break again... unwrap feels dangerous here.

### DIFF
--- a/src/page.rs
+++ b/src/page.rs
@@ -182,10 +182,20 @@ impl<T: serde::de::DeserializeOwned> crate::FromResponse for Page<T> {
                 last,
             })
         } else {
-            let attr = vec!["items", "workflows", "workflow_runs", "jobs", "artifacts", "repositories"]
-                .into_iter()
-                .find(|v| json.get(v).is_some())
-                .unwrap();
+            let attr = vec![
+                "items",
+                "workflows",
+                "workflow_runs",
+                "jobs",
+                "artifacts",
+                "repositories",
+            ]
+            .into_iter()
+            .find(|v| json.get(v).is_some())
+            .ok_or(Box::from(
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your issue! Do you want to change your PR to replacing the replace unwrap with a `.map_err` to an error variant from `octocrab::Error` instead? That would actually fix the panic issue.

gabm:
I will have a look into this, yes

gabm:
honestly, I am not able to use snafu properly here :( my attempt is:

```rust
            let attr = vec![
                "items",
                "workflows",
                "workflow_runs",
                "jobs",
                "artifacts",
                "repositories",
            ]
            .into_iter()
            .find(|v| json.get(v).is_some())
            .ok_or(crate::error::OtherSnafu)?;
```

but the `?` causes problems here: with `^ the trait `std::convert::From<OtherSnafu>` is not implemented for `error::Error``

XAMPPRocky:
@gabm You need to call it with `snafu::Context` e.g.

```rust
    .context(crate::error::OtherSnafu { source: Box::from("error message") })
```

gabm:
I got it done :+1:

XAMPPRocky:

PR #283, opened 2022-12-16
Make Review::user optional

MarcoFalke:
According to the docs, it is optional:

```
      "user": {
        "anyOf": [
          {
            "type": "null"
          },
          {
            "title": "Simple User",
            "description": "A GitHub user.",
            "type": "object",
            "properties": {
...
```

From https://docs.github.com/en/rest/pulls/reviews?apiVersion=2022-11-28#list-reviews-for-a-pull-request

This should fix crashes of the form:

```
thread '' panicked at 'called `Result::unwrap()` on an `Err` value: Serde {
 source: Error("invalid type: null, expected struct User", line: 0, column: 0), backtrace: Backtrace(
   0: <snafu::backtrace_shim::Backtrace as snafu::GenerateImplicitData>::generate

### DIFF
--- a/src/models/pulls.rs
+++ b/src/models/pulls.rs
@@ -197,7 +197,7 @@ pub struct Review {
     pub id: ReviewId,
     pub node_id: String,
     pub html_url: Url,
-    pub user: User,
+    pub user: Option<User>,
     #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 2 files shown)


aureleoules (approved):
Looks good to me.

XAMPPRocky:
Thank you for your PR!

PR #284, opened 2022-12-17
Fix type Vec<Repositories> to InstallationRepositories on example

riseshia:
## Summary

example codes aren't working, so let them fixed.

## Detail

`/installation/repositories` returns repositories with total count.
[API Reference](https://docs.github.com/en/rest/apps/installations?apiVersion=2022-11-28#list-repositories-accessible-to-the-app-installation)
(maybe it had returned just repositories? 🤔 )

octocrab has `InstallationRepositories` type, however, these examples has wrong type. ;)
I confirms It works with this fix.


XAMPPRocky:

PR #286, opened 2022-12-23
CD Workflow

lineville:
Adds a release workflow to cut a new release automatically on pushes to main or with a manual release. @XAMPPRocky will need to set the repo secret `CARGO_REGISTRY_TOKEN` for this to work. Hopefully this will help to increase development velocity on the project and reduce the time/effort of cutting manual releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,22 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - main
+  release:
+    types: [published]
+
+jobs:
+  publish-cargo:
+    name: Publish
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - uses: katyo/publish-crates@v1
+        with:
+          registry-token: ${{ secrets.CARGO_REGISTRY_TOKEN }}


XAMPPRocky:
Thank you for your PR! I've been planning on adding the same thing for awhile now. This GH action looks interesting however it's lacking some of the features I want, because I want this not only for octocrab but all of my projects.

The main features missing are semver checks, CHANGELOG updates, and rolling releases. I'd like to have it so the GitHub action detects when there's public API breaking changes and make the appropriate bump in the version number. Secondly I don't want to release on every merge, because sometimes, people get stuff wrong or something is missed in a review. So I'd like to have a grace period where the code can be updated before it's released, so that there isn't an immediate pressure to release a new version, I instead want to have it so every week there is a new version that releases all the changes that are at least a week old.

I'll try to work on this soon, as I recognise the demand for more timely releases.

lineville:
I think that you should be able to add workflows to your personal `.github` repo and then those workflows can be used across multiple projects, although you'd probably want different varients of publishing workflows for different projects. I'm not aware of any tool that will intelligently detect "breaking changes" and make the appropriate semver adjustment, but a nice happy medium would be to just trigger this workflow on `release: [published]` and remove the trigger on merges to main. Then you would have full control over the published version by setting it manually when creating a release on GitHub, then the publish to crates.io would be handled automatically. Totally understand wanting to maintain some more control and not just cut releases on every merge, thanks for the work you're doing here ❤️

XAMPPRocky:
> I'm not aware of any tool that will intelligently detect "breaking changes" and make the appropriate semver adjustment

Someone's already made it for Rust projects. It's called `cargo-semver-checks`, so making that adjustment is as simple as calling that, and if it is breaking, set a major bump using a tool like `cargo-workspaces`.

lineville:
> 

Ah nice, good to know 👍

MarcoIeni:
Hey! I created a Rust project that handles both releases and changelog updates. You can check it out [here](https://marcoieni.github.io/release-plz/).
It also allows the maintainer to decide when to release (by merging a pr).
Could this work for you?

Also, if you want to collaborate, there's an [open issue](https://github.com/MarcoIeni/release-plz/issues/373) to integrate `cargo-semver-checks` into it!

XAMPPRocky:
@MarcoIeni Thank you for your suggestion Yes, I am interested in your project, it does seem to do most of what I want to achieve. I'd love to collaborate on adding the feature, though right now my time is currently quite limited with other projects, but if no one else contributes to it, I'd hope to be able to help add it within the next couple months. The other question I would have, not for octocrab but my other projects is how well does it handle projects that are libraries and binaries? I'd like to have also use this for tokei.

---

In the meantime I've opened #293 to track this.

MarcoIeni:
> how well does it handle projects that are libraries and binaries?

At the moment, release-plz treats libraries and binaries in the same way.

For the future, I have two issues opened for binaries:
- https://github.com/MarcoIeni/release-plz/issues/48
- https://github.com/MarcoIeni/release-plz/issues/419

Do you have something specific in mind? 🤔

XAMPPRocky:
Closing as superseded by release-plz integration. Thank you to @lineville for kicking this off!

PR #291, opened 2023-01-08
feat(releases): add make-latest field

MarcoIeni:


XAMPPRocky:
Thank you for your PR!

src/api/repos/releases.rs
```diff
@@ -265,6 +265,28 @@ pub struct CreateReleaseBuilder<'octo, 'repos, 'handler, 'tag_name, 'target_comm
...
+pub enum MakeLatest {
+    True,
+    False,
+    Legacy,
+}
+
+impl std::fmt::Display for MakeLatest {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        let text = match self {
+            Self::False => "false",
```
sassman, line 283: FYI: just in case you are not aware, but you can delegate this "manual" string mapping to serde, since you use already the `rename_all = "lowercase"` attribute, serde will take care of renaming the enum variants accordingly.

You can find an example [here](https://github.com/curlz-rs/curlz/blob/main/src/curlz/data/request/http_method.rs#L21) and [a test here](https://github.com/curlz-rs/curlz/blob/main/src/curlz/data/request/http_method.rs#L52)

MarcoIeni: Thanks, I suspected there was a better way to do it :)

XAMPPRocky: Of course while that solves the problem for serde, it doesn't let you easily print the value in other contexts, I don't think you should have go through a whole serialisation process just to print the value, so I think having the display impl is good regardless of serde.

PR #294, opened 2023-01-13
Replace `reqwest` with `http` and `hyper`.

### LINKED ISSUE #99: Replace `reqwest` with `http`
In order to be more client agnostic, and allow using Octocrab in more places, I would like to remove `reqwest` and replace it entirely with `http`, `tower`, and `hyper`. Using the same approach as `kube-rs`.

If anyone is interested in taking this on, I would make a PR per endpoint or endpoint group, and not move all of them at once. Feel to ask on the issue or in the [discussion board](https://github.com/XAMPPRocky/octocrab/discussions) if you need help.

IgnisDa:
Fixes #99.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -160,11 +160,12 @@ pub mod params;
 use std::fmt;
 use std::sync::{Arc, RwLock};
 
+use http::{header::HeaderName, StatusCode};
 use once_cell::sync::Lazy;
-use reqwest::{header::HeaderName, StatusCode, Url};
 use secrecy::{ExposeSecret, SecretString};
 use serde::Serialize;
 use snafu::*;
+use url::Url;
 
 use auth::{AppAuth, Auth};
 use models::{AppId, InstallationId, InstallationToken};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,10 +93,12 @@
 //! the request.
 //!
 //! ```no_run
+//! # use std::str::FromStr;
+//! # use url::Url;
 //! # async fn run() -> octocrab::Result<()> {
 //! let octocrab = octocrab::instance();
 //! let response =  octocrab
-//!     ._get("https://api.github.com/organizations", None::<&()>)
+//!     ._get(Url::from_str("https://api.github.com/organizations").unwrap(), None::<&()>)
 //!     .await?;
 //!
 //! // You can also use `Octocrab::absolute_url` if you want to still to go to
@@ -158,6 +160,7 @@ pub mod models;
 pub mod params;
 
 use std::fmt;
+use std::str::FromStr;
 use std::sync::{Arc, RwLock};
 
 use http::{header::HeaderName, StatusCode};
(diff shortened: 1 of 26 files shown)


src/lib.rs
```diff
@@ -313,8 +317,8 @@ impl OctocrabBuilder {
     }
 
     /// Set the base url for `Octocrab`.
-    pub fn base_url(mut self, base_url: impl reqwest::IntoUrl) -> Result<Self> {
-        self.base_url = Some(base_url.into_url().context(crate::error::HttpSnafu)?);
+    pub fn base_url(mut self, base_url: &str) -> Result<Self> {
```
XAMPPRocky, line 320: This should be `AsRef<str>`, same with the other methods.

IgnisDa: Done! I am fairly new to rust so I might have these amateurish errors 🥲.

src/lib.rs
```diff
@@ -313,28 +317,28 @@ impl OctocrabBuilder {
     }
 
     /// Set the base url for `Octocrab`.
-    pub fn base_url(mut self, base_url: impl reqwest::IntoUrl) -> Result<Self> {
-        self.base_url = Some(base_url.into_url().context(crate::error::HttpSnafu)?);
```
XAMPPRocky, line 317: This should use `TryFrom<Url>` instead of `AsRef`, because then we can still accept `Url` in addition to `&str`.

src/lib.rs
XAMPPRocky: Everyone has to learn sometime, don't be afraid to ask questions in the PR if you're unsure about something 🙂

src/lib.rs
IgnisDa: Looking at the docs: https://doc.rust-lang.org/std/convert/trait.TryInto.html
> Library authors should usually not directly implement this trait, but should prefer implementing the [TryFrom](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) trait, which offers greater flexibility and provides an equivalent TryInto implementation for free, thanks to a blanket implementation in the standard library.

IgnisDa: >  but you can still use TryInto where appropriate

Can you tell the syntax to this please?

XAMPPRocky: It's the same as everything else.
```rust
fn foo(url: impl TryInto<Url>) -> Result<()> {
     let url = url.try_into()?;
     Ok(())
}
```

L1ghtman2k:
What a coincidence, we were actually just getting started with implementing tower/hyper/http for octocrab. I will likely make a separate PR for this

IgnisDa:
@L1ghtman2k If you'd like, you can edit this PR. I don't think I'll be able to do the hyper integration; it's too complicated for me. I'm more than happy to add you as a contributor to my form.

src/lib.rs
IgnisDa: I tried to use it, however I could not get the types to match up.

L1ghtman2k:
> @L1ghtman2k If you'd like, you can edit this PR. I don't think I'll be able to do the hyper integration; it's too complicated for me. I'm more than happy to add you as a contributor to my fork.

Hey, it is definitely ambitious, I am myself am relatively new to rust, will see how far I can get.  No need to give access, I have already cherry picked some of your changes

XAMPPRocky:
The hyper integration should be straightforward, it's replacing `reqwest::Client` with a dynamic `tower::Service`, look at the `kube` crate for inspiration. https://github.com/kube-rs/kube/blob/ce69d9238c4dcd65b50f7647f485922169ab078d/kube-client/src/client/mod.rs#L65

IgnisDa:
@L1ghtman2k Any progress made on this?

IgnisDa:
Awesome! Closing this in favor of #297.

Issue #175, opened 2021-12-28
Conditional Requests / HTTP Caching


BGR360:
The GitHub API supports [conditional requests per HTTP RFC 7234](https://httpwg.org/specs/rfc7234.html). From the [GitHub API docs](https://docs.github.com/en/rest/guides/getting-started-with-the-rest-api#conditional-requests) (emphasis my own):

> A big part of being a good API citizen is respecting rate limits by caching information that hasn't changed. The API supports conditional requests and helps you do the right thing.
> [...]
> The 304 status indicates that the resource hasn't changed since the last time we asked for it and the response will contain no body. **As a bonus, 304 responses don't count against your rate limit.**

For those that have ambitious goals for their GitHub-integrated projects, this is very valuable, as the 5,000-requests-per-hour rate limit may become very limiting without caching.

Another crate, [`octorust`](https://crates.io/crates/octorust), provides support for HTTP caching as an optional feature. I was planning to use `octorust` for a project for this reason, but I discovered Octocrab and I would prefer to use the higher-level constructs provided by it.

I think the path to supporting this may not be too hard. There is a [`reqwest-middleware-cache`](https://crates.io/crates/reqwest-middleware-cache) crate that appears to provide a very easy shim layer around a `reqwest::Client`.

Here's what I imagine the usage would look like:

Cargo.toml:

```toml
[dependencies]
octocrab = { version = "^0.15", features = ["http_cache"] }
```

main.rs:

```rust
let octocrab = octocrab::Octocrab::builder()
    .http_cache("~/.cache/github")
    .build()?;
```

XAMPPRocky:
Thank you for your issue! While I would definitely like to have this feature in octocrab, I'm hesitant to add any reqwest middleware as I want to remove reqwest as a dependency in this library (see #168), and adding middleware would make this harder.

Also I would not want to just have filesystem based caching, ideally it would be a pluggable solution (such as a trait) that allows people to provide their own caching implementation, because a lot of people aren't having long running processes and don't want to fill the filesystem with cache requests when we're also not providing cleanup.

So to summarise, I'm in favour of having it, but want to move to Sans-IO first (or use an agnostic approach), and I want a cache storage trait that handles the actual caching problem.

iwahbe:
Has there been any progress on this issue? If not, I'd be interested in contributing. I don't want to duplicate existing work though.

XAMPPRocky:
[...]

PR #297, opened 2023-01-16
Replace `reqwes`t with `http` and `hyper`, and `tower`

L1ghtman2k:
Fixes https://github.com/XAMPPRocky/octocrab/issues/99


L1ghtman2k:
Hey @XAMPPRocky , this is still a WIP and needs a lot more in refactoring and cleaning up, I am just pushing this here to indicate that it is being worked on.
I also need to work on OctocrabBuilder to layer proper/required services, and expose properties that could be used by clients.

But thus far, I have a question:
 - Does it makes sense to move base_uri to a service layer(that is how I implemented it in current PR, and that's how kube-rs implements it), instead of having it be constructed it every time on the fly? i could see a few pros for this, like how it makes all caller functions a lot cleaner(since you wouldn't need to interpolate strings with base_uri). However, it does introduce a niche(IMHO) issue with cloning of octocrab client. (You wouldn't be able to change base_uri of the cloned client, at least not easily)

XAMPPRocky:
> You wouldn't be able to change base_uri of the cloned client(since it will clone underlying service that was already constructed)

I'd need to see the service code in kube and this more, but just from your description. My initial thought for the solution to this problem is to have the service wrapped in a `ArcSwap`, this way we can update the underlying service, while still having all the shallow clone benefits of `Arc`.

L1ghtman2k:
Service in Kube does it via https://docs.rs/tower/latest/tower/buffer/struct.Buffer.html, I am wondering if tower has some out of the box solution for this

L1ghtman2k:
another potential improvement is to replace url lib completely with hyper::Uri, and hyper_serde crate, though, it does become a little redundant to annotate Uri fields with 
```
#[serde(deserialize_with = "hyper_serde::deserialize",
serialize_with = "hyper_serde::serialize")] 
```
annotation... I think for now it is better to keep the URL crate for serialization/deserialization operations. Plus, I am not entirely sure how hyper_serde behaves when uri is wrapped around constructs like Option

L1ghtman2k:
Also, could you re-approve the workflow rerun? I think all works on my end, just want to make sure it compiles in CI

L1ghtman2k:
I worked around the base URI service issue by making it the outer layer, which wraps BufferService inside. This essentially means that we are only full cloning the base_uri service when making .clone() call, everything else is a shallow copy offered by BufferService

L1ghtman2k:
At this point, I think this is almost done. The few things that are left are: implement some retry layer(hidden behind retry feature, just like timeout), and fix docs

L1ghtman2k:
A few things I am not sure what to decide on:

Currently, the build() function of OctocrabBuilder implements various hyper-client configurations (tls, timeout), and additional layers(tls, retry, extra-headers)  via features that are enabled by default. In addition, OctocrabBuilder contains a few fields that configure some of the settings for these functionalities(e.g. connection write timeout).

This is a rather opinionated approach and leaves little room for customization in terms of adding your own layers directly into the builder.

If the end user would like to introduce some additional layer (say metrics layer before, or after retry), or replace the current implementation with their own (e.g. custom tracing), or caching(https://github.com/XAMPPRocky/octocrab/issues/175), they would have to use Octocrab::new() and construct their own service from scratch, which could lead to some issues if developers not careful(e.g. while Octocrab::new does take care of adding BaseUri layer, it doesn't ensure that service contains necessary layers, like ExtraHeaders for auth).

We could leave it as is, we would just have to provide some good examples(or reference the OctocrabBuilder::build function)

Another way we could go about it is to make the Octocrab Builder completely unopinionated, and let users specify their own layers from scratch(kube-rs way):
```
    pub fn with_layer<L: Layer<T>>(layer: L) -> Self<L::Service> {
        Self {
            service_builder: ServiceBuilder::new().layer(layer),
            ..Self
        }
    }
```

This has a plus since it makes OctoCrab reusable for custom approaches. However, this makes the configuration slightly more confusing for users who aren't too interested in customizing their services, and leaves a lot of room for potential pitfalls(forgetting ExtraHeader Layer for auth, not enabling tls, etc)


There are approaches that could be implemented by combining both:
 - Move important pieces (we would have to decide which ones, I think at the very least ExtraHeader's auth portion) to Octocrab::new(), just like base_uri, and then let clients use `with_layer` any way they want.
 - Supporting "best practices" layers(e.g. retry, tls, tracing) with opinionated implementation. One way to implement this would be to create an opinionated `with_layer_X`(ex: with_layer_timeout(write_timeout, read_timeout ...)

Currently, I am personally leaning more towards current implementation, and backing it up with better examples on how to construct custom clients

XAMPPRocky:
Thank you again for your continued work on this! 🙏

My thoughts are we should do both. 🙂 We should move those components to layers, and have the default setup construct an opinionated set of layers that you can build ontop of with `with_layer`, and then additionally have an alternative constructor for the builder that doesn't include any of those layers, so you can customise it, and the layers we add are exposed in a public`layers` module, so if you want to customise it, you still have access to opt back into some of the layers we provide.

L1ghtman2k:
Hey @XAMPPRocky, I attempted to create a builder that can work for the service-based client.

In short, it allows the end user to use Octocrab Builder via the `DefaultOctocrabBuilderConfig` config, which implements all opinionated best practices and allows minor configuration of those options.

In addition, the end users can either use bare-bones with_layer implementation, or they can implement their own implementation of Octocrab Builder, provided they base it on a `Config`

Let me know your thoughts on this

L1ghtman2k:
One thing that's worth mentioning is that functions like `fn get` will now use TryUri instead of reqwest::IntoUrl. There are a few behaviors that are different, for example, in case of TryUri routes must be prefixed with `/`, otherwise it might be interpreted unexpectedly: 

`/installations/orgs` is interpreted as:
```
scheme: None
authority: None
Path: /installations/orgs
```

while `installations/orgs` is interpreted as:
```
scheme: None
authority: installations
Path: `orgs`
```

XAMPPRocky:
> I am not exactly sure best way to notify users about this change.

I would just add an entry to the changelog.

I hope to have time to review this soon!

IgnisDa:
Hey @XAMPPRocky any updates on this?

L1ghtman2k:
Hey @goodspark, I was just synced fork with the master, and I saw you have implemented tracing and configurable retries.

I have actually added them as hyper layers, and I was wondering if you could look over the parts of my PR related to tracing and retries, and comment on if it might be missing anything... 

Tracing Layer:
https://github.com/XAMPPRocky/octocrab/pull/297/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R551

Retry Layer(default 3, but can be changed):
https://github.com/XAMPPRocky/octocrab/pull/297/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R512

XAMPPRocky:
I'll review this PR later this month with a release soon afterwards

L1ghtman2k:
(I am still merging few things, so it will fail rn, thanks for heads up!)

src/service/middleware/retry/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+            RetryConfig::None => None,
+            RetryConfig::Simple(count) => match result {
+                Ok(_) => None,
+                Err(_) => {
+                    if *count > 0 {
+                        Some(future::ready(RetryConfig::Simple(count - 1)))
+                    } else {
+                        None
+                    }
+                }
```
goodspark, line 30: Generally retries should only occur when the response status code is 5xx, 429, and (in the odd case of Github) 403s (because they chose to use that code to indicate rate-limit issues instead of 429s). Other 4xx codes generally mean the client/user did something wrong and sending the same exact request won't solve the problem. Whereas with 5xxs (server error), waiting a bit and trying again might help. And obviously waiting and trying the same request can help for rate-limit issues.

We might want to hardcode that into the client _or_ provide a predicate for library users to provide so they can provide their own logic.

src/service/middleware/retry/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+                let mut new_req = Request::builder()
+                    .uri(req.uri())
+                    .method(req.method())
+                    .version(req.version());
+                for (name, value) in req.headers() {
+                    new_req = new_req.header(name, value);
+                }
+
+                let body = req.body().clone();
+                let new_req = new_req.body(body).expect("failed to build request");
```
goodspark, line 49: Personally, I'd prefer libraries _not_ panic and instead return errors and let upstream users decide if they want to panic or not (although in the case, I think [Tower will just not retry if we return None](https://docs.rs/tower/latest/src/tower/retry/future.rs.html#78-88)). But I don't know the philosophy on this at the crate level.

src/service/middleware/retry/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+
+                Some(new_req)
+            }
+        }
+    }
+}
```
goodspark, line 55: I can't find anything about tracing and the link doesn't seem to point to any specific file. Was it removed from this PR?

src/service/middleware/retry/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+            RetryConfig::Simple(count) => match result {
+                Ok(_) => None,
+                Err(_) => {
+                    if *count > 0 {
+                        Some(future::ready(RetryConfig::Simple(count - 1)))
+                    } else {
+                        None
+                    }
+                }
+            },
```
goodspark, line 31: It'd be nice, but not strictly a blocker, if we could somehow pass the retry attempt number to the tracing layer so we can include it as a field (`http.resend_count`) on the API call span. Ref: https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#http-client

src/service/middleware/retry/mod.rs
L1ghtman2k: lib.rs, line 550. I think hyperlink doesn't work because lib.rs has too many changes

src/service/middleware/retry/mod.rs
L1ghtman2k: good point. I am wondering what the best way would be to pass that information down? Currently retry layer wraps tracing layer. We could just modify the request header to add the retry attempt, and then have it removed on tracing layer(seems kinda ugly to me)? Alternatively, we can have tracing layer wrap retry layer, and let retry layer create its own span with `http.resend_count`


Also, it seems like there is a rename planned for `http.resend_count`, so we might want to have this implemented after?: 
https://github.com/open-telemetry/opentelemetry-specification/issues/3182

src/service/middleware/retry/mod.rs
L1ghtman2k: The reason it has `expect` there is because it should never actually panic there (we are reconstructing Request<String> by cloning components of &Request<String>, so there shouldn't be any invalid values/headers that would cause panic). I will try to clarify with a comment

src/service/middleware/retry/mod.rs
L1ghtman2k: Code above currently only retries when unexpected http requests come in, for instance, EOF, or some other HTTP issues. I am a little hesitant about 403s?

I just added logic to handle 5**, and 429.

BTW tower approach does allow clients to create more complex/opinionated retry logic if they want and replace our implementation.

XAMPPRocky:
I've reformatted the project as I noticed a few of the changes were just format changes, so you'll need to rebase it to fix the conflicts.

L1ghtman2k:
This PR does cargo fmt. Could we use the one from this PR instead?

XAMPPRocky:
@L1ghtman2k No because it makes this PR harder to review if there's unrelated style changes.

L1ghtman2k:
> @L1ghtman2k No because it makes this PR harder to review if there's unrelated style changes.

Sounds good, I just got around to resolve the conflicts

L1ghtman2k:
Wired, GitHub UI merge conflict resolution is a little different than my IDEs "Sync Fork"..., merge conflicts should be resolved now...?

Cargo.toml
```diff
@@ -28,35 +27,41 @@ hyperx = "1.3.0"
 snafu = { version = "0.7", features = ["backtraces"] }
 once_cell = "1.7.2"
 arc-swap = "1.3.0"
-base64 = "0.20.0"
+base64 = "0.13.1"
```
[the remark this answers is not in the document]
L1ghtman2k: I don't think that is the case? (maybe the comment might be outdated?)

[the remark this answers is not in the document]
L1ghtman2k: I might have downgraded at some point in one of the oldest commits when I was testing compatibilities, but it should not be changing the 0.21.0 version that is on master branch as of latest commit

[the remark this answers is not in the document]
XAMPPRocky: Yeah sorry I believe this was from an earlier review but I hit add to review rather than single comment

XAMPPRocky:

PR #242, opened 2022-09-11
Add PullRequestReviewEvent

sagudev:
GitHub docs: https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request_review

PR #305, opened 2023-02-12
Fix PullRequestReviewEvent model

LuaKT:
The previous PR to add this event (#242) incorrectly used the webhook event schema. This fixes that to use the event api schema: https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#pullrequestreviewevent


XAMPPRocky:
Thank you for your PR!

PR #306, opened 2023-02-23
Allow customized retry decisions

goodspark:
Adds a retry predicate (with defaults to match existing behavior for 401s). 5xx codes are generally worth retrying without changing anything about the request and this will let any user customize the behavior.


XAMPPRocky:

PR #307, opened 2023-02-24
Add tracing to requests

goodspark:
Make it a feature so people can opt-in


src/lib.rs
```diff
@@ -920,10 +925,40 @@ impl Octocrab {
...
+                error=field::Empty,
+            );
+            let result = if cfg!(feature = "trace") {
+                let service = self.base_url.host_str().unwrap_or(GITHUB_SERVICE);
+                span.record("service.name", service);
+                span.in_scope(|| {
+                    request.send()
+                }).await
+            } else {
+                request.send().await
```
XAMPPRocky, line 944: Shouldn't these `await`s be instrumented with the span? Otherwise it could be inaccurate I believe from the last time I read the docs.

Cargo.toml
```diff
@@ -48,6 +49,7 @@ wiremock = "0.5.3"
 [features]
 default = ["native-tls"]
 rustls = ["rustls-tls"]  # Leagcy support (<=0.17.0)
+trace = ["tracing"]
```
XAMPPRocky, line 52: This shouldn't be needed as by default making tracing optional creates a feature called `tracing`.
```suggestion
```

src/lib.rs
```diff
@@ -893,6 +899,38 @@ impl Octocrab {
         }
     }
 
+    #[cfg(feature = "trace")]
```
XAMPPRocky, line 902:
```suggestion
    #[cfg(feature = "tracing")]
```

src/lib.rs
```diff
@@ -179,6 +179,12 @@ pub use self::{
     page::Page,
 };
 
+#[cfg(feature = "trace")]
+use tracing::{self, field};
+
+#[cfg(feature = "trace")]
```
XAMPPRocky, line 185:
```suggestion
#[cfg(feature = "tracing")]
use tracing::{self, field};

#[cfg(feature = "tracing")]
```

src/lib.rs
```diff
@@ -893,6 +899,38 @@ impl Octocrab {
...
+            },
+            Err(e) => {
+                let status = e.status().and_then(|s| Some(s.as_u16()));
+                span.record("status_code", status);
+            },
+        };
+        result
+    }
+
+    #[cfg(not(feature = "trace"))]
```
XAMPPRocky, line 929:
```suggestion
    #[cfg(not(feature = "tracing"))]
```

Cargo.toml
goodspark: I decided to just remove the feature and make it part of the library. Would that be OK?

src/lib.rs
```diff
@@ -893,6 +897,39 @@ impl Octocrab {
...
+            Ok(v) => {
+                span.record("http.status_code", v.status().as_u16());
+            }
+            Err(e) => {
+                let status = e.status().and_then(|s| Some(s.as_u16()));
+                span.record("http.status_code", status);
+            }
+        };
+        result
+    }
```
goodspark, line 931: Had to pull things out to a sep async function because `span` would've moved if I used `.instrument(span)`, which then prevents recording fields after the request.

goodspark:
Also, sorry about the few force pushes. Thought I still had some time to do some last minute tweaks.

XAMPPRocky:
Thank you for your PR!

PR #308, opened 2023-02-25
Add target URL to Status model

goodspark:
Schema: https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#about-commit-statuses


XAMPPRocky:
Thank you for your PR!

PR #313, opened 2023-03-10
Add `client_builder` option to `OctocrabBuilder`

MTCoster:
This allows custom options to be set on the `reqwest::Client` used by Octocrab.

My motivating use case here is the use of an internal CA on an enterprise GitHub instance.

Unless I can call [`ClientBuilder::add_root_certificate()`](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.add_root_certificate), I end up with:

```
Error: HTTP Error: error sending request for url (<redacted>): error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -279,6 +279,7 @@ pub struct OctocrabBuilder {
     extra_headers: Vec<(HeaderName, String)>,
     base_url: Option<Url>,
     retry_predicate: Option<RetryPredicate>,
+    client_builder: Option<reqwest::ClientBuilder>,
 }
 
 impl OctocrabBuilder {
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! However reqwest is going to be removed with introduction of #297. So I'm hesitant to accept the PR only to later scrap it.

You should be able to achieve the same thing with that PR.

MTCoster:
That works for me! I wasn't sure how long it would take to land #297, so I thought a smaller change would have a better chance of making it in sooner. I don't mind waiting though :)

XAMPPRocky:
Closing since #297 has been merged

PR #323, opened 2023-04-01
Percent encode label name in remove_label to avoid InvalidUriChar error

MarcoFalke:


XAMPPRocky:
Thank you for your PR!

PR #325, opened 2023-04-05
Added poll org events

Klaven:
I was looking to use octocrab but I was looking to pull the events at an org level and noticed you had only implemented it on the repo level.

This PR looks bigger then it is, as I mostly just used what you had done with the repo events. IMO it felt like there should be a better way to do this that copies less code but it did not jump out at me. Pretty new to your api so please let me know if there was a better way to do this and I'll do it!

Another thing is I did not know how we might test this feature. If you all have a way that you are doing that I'm happy to oblige.

Thank you!

### DIFF
--- a/examples/poll_org_events.rs
+++ b/examples/poll_org_events.rs
@@ -1,0 +1,41 @@
+use octocrab::{etag::Etagged, models::events::Event, Page};
+use std::collections::VecDeque;
+
+const DELAY_MS: u64 = 500;
+const TRACKING_CAPACITY: usize = 20;
+
+#[tokio::main]
+async fn main() -> octocrab::Result<()> {
+    let mut etag = None;
+    let mut seen = VecDeque::with_capacity(TRACKING_CAPACITY);
+    let octo = octocrab::instance();
+    loop {
+        let response: Etagged<Page<Event>> = octo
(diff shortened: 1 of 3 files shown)


Klaven:
I was thinking that we could make a single `ListEventsBuilder` that took a `EventHandler` trait that both `OrgHandler` and `RepoHandler` could implement maybe? If this or something like this is the solution i'm happy to do that as a follow up PR!

XAMPPRocky:

> If this or something like this is the solution i'm happy to do that as a follow up PR!

If it can be easily shared then it's more than welcome to reduce the amount of boiler plate, are there other APIs that would benefit from a generic events handler?

Klaven:
> Thank you for your PR, and congrats on your first contribution! tada
> 
> > If this or something like this is the solution i'm happy to do that as a follow up PR!
> 
> If it can be easily shared then it's more than welcome to reduce the amount of boiler plate, are there other APIs that would benefit from a generic events handler?

Not sure, I'll look and see, there are a few "events" I'll dig though the github docs and what you have done so far and see if it's "just here" or if there are other places as well. If it is more then just here I'll see what I can do to reduce the boiler plate for these types of calls.

Issue #27, opened 2020-07-11
Add Gists API


XAMPPRocky:

envp:
@XAMPPRocky is there still interest in reviewing a contribution for this?

XAMPPRocky:
@envp Contibutions are always welcome and encouraged 🙂

envp:
> @envp Contibutions are always welcome and encouraged 🙂

Thanks! I've made a couple of PRs to familiarize myself.

envp:
I've tried to tabulate what the `TODO`s would be acc to [GitHub API docs (API Version 2022-11-28)][1]

- [X] Create a gist (`POST /gists`)
- [X] Get a gist (`GET /gists/{gist_id}`)
- [X] Update a gist (`PATCH /gists/{gist_id}`)
- [X] List gist commits (`GET /gists/{gist_id}/commits`)
- [X] List gists for authenticated user (`GET /gists`)
- [ ] List all public gists (`GET /gists`)
- [ ] List public gists (`GET /gists/public`)
- [ ] List starred gists (`GET /gists/starred`)
- [ ] Delete a gist (`DELETE /gists/{gist_id}`)
- [ ] List gist forks (`GET /gists/{gist_id}/forks`)
- [ ] Fork a gist (`POST /gists/{gist_id}/forks`)
- [ ] Check if a gist is starred (`GET /gists/{gist_id}/star`)
- [ ] Star a gist (`PUT /gists/{gist_id}/star`)
- [ ] Unstar a gist (`DELETE /gists/{gist_id}/star`)
- [ ] Get a gist revision (`GET /gists/{gist_id}/{sha}`)
- [ ] List gists for a user (`GET /users/{username}/gists`)


[1]: https://docs.github.com/en/rest/gists/gists?apiVersion=2022-11-28#create-a-gist

let me know if I missed something.

envp:
@XAMPPRocky Some of github's APIs return an empty response body, these can't be deserialized with the current `FromResponse` trait which assume there's always valid JSON.

I tried implementing `FromResponse for ()` so callers can ask to discard the response body. However, that conflicts with the current implementation.  WDYT is a good way to handle this?

XAMPPRocky:
@envp There are other parts of the API that do that, in general for those you should use the `_` prefixed versions like `_put` that do no processing on the response.

envp:
[...]

PR #328, opened 2023-04-08
current: Add `CurrentAuthHandler::list_gists_for_authenticated_user`

envp:
This PR partially addresses #27 through the following:

* Extend `CurrentAuthHandler` by allowing pagination through the authenticated user's gists
* Bugfix the `Gist` model which incorrectly assumed that `content` and `truncated` fields are always present
* Add an example demonstrating use of the new function. [Sample output](https://gist.github.com/envp/e021fc416b51016591386304ed3d4d4c)


XAMPPRocky:
Thank you for your PR!

PR #329, opened 2023-04-08
examples: Add an example showing gist creation

envp:
This takes the example from [`CreateGistHandler`'s rustdoc](https://github.com/XAMPPRocky/octocrab/blob/775936c7731b6a8cae4f0a2d4628d1d03e341e24/src/api/gists.rs#L25-L43) and adds it as a runnable example. 

Links to: #27


XAMPPRocky:
Thank you for your PR!

PR #330, opened 2023-04-09
gists+examples: Extend `GistsHandler` through `star(...)`, `unstar(...)`, `is_starred(...)`

envp:
Adding code related to `/gists/{gist_id}/star` here.

TODO:

- [X] Star a gist (`PUT /gists/{gist_id}/star`)
- [x] Unstar a gist (`DELETE /gists/{gist_id}/star`)
- [x] Check if a gist is starred (`GET /gists/{gist_id}/star`)

Related to: #27


envp:
cc: @XAMPPRocky this is ready for review. Thanks for your patience with the PR spam :^)

examples/star_unstar_a_gist.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    let gist_id = if let Some(gist_id) = std::env::args().nth(2) {
+        gist_id
+    } else {
+        eprintln!("error: Need GIST_ID as second argument.");
+        eprintln!("{}", USAGE);
+        std::process::exit(1);
+    };
+    ProgramArguments { gist_id, star }
+}
```
envp, line 37: I've added this helper since I wasn't sure if pulling in a CLI dependency for a couple of arguments was a good idea. I don't have any strong opinions on these 3 endpoints living in the same example, happy to split them if this is considered too complex for an illustrative example.

src/api/gists.rs
```diff
@@ -122,6 +123,79 @@ impl<'octo> GistsHandler<'octo> {
...
+    ///
+    /// [docs]: https://docs.github.com/en/rest/gists/gists?apiVersion=2022-11-28#check-if-a-gist-is-starred
+    pub async fn is_starred(&self, gist_id: impl AsRef<str>) -> Result<bool> {
+        let gist_id = gist_id.as_ref();
+        let response = self
+            .crab
+            ._get(format!("/gists/{gist_id}/star"))
+            .await?;
+        // Gist API returns 204 (NO CONTENT) if a gist is starred
+        Ok(response.status() == StatusCode::NO_CONTENT)
```
envp, line 149: Do we need a `StarGistHandler` to wrap the http specifics?

XAMPPRocky: I'm not sure what you mean, what other semantics?

examples/star_unstar_a_gist.rs
XAMPPRocky: I don't have a strong opinion on examples as long they're clear, I also don't have a problem with using dependencies in them either.

The point of an example is to provide an minimal easily understandable version, anything that helps with that is encouraged.

XAMPPRocky:
@envp Thank you for your PR! Spam would imply that it's unwelcome or not wanted, which I can assure you, your contributions are not. 🙂

examples/star_unstar_a_gist.rs
envp: Thanks, noted. I'll keep that in mind for future examples!

src/api/gists.rs
envp: Ah, I meant I wasn't sure if you preferred hiding the http status check behind a builder. I'm not sure what the "correct" approach was so I went with the simplest one.

XAMPPRocky: No, a builder should only be used if there's optional parameters, if there's only required parameters (or no parameters), then a builder is not needed.

XAMPPRocky:
Thank you for your PR!

PR #331, opened 2023-04-10
Add graphql method that passes variables with the payload

Libbum:


Libbum:
(sorry, this wasn't ready)

PR #333, opened 2023-04-10
Handle `DELETE /gists/{gist_id}`

envp:
Relates to: #27

### DIFF
--- a/src/api/gists.rs
+++ b/src/api/gists.rs
@@ -79,6 +79,22 @@ impl<'octo> GistsHandler<'octo> {
         self.crab.get(format!("/gists/{id}"), None::<&()>).await
     }
 
+    /// Delete a single gist.
+    ///
+    /// ```no_run
+    /// # async fn run() -> octocrab::Result<()> {
+    /// octocrab::instance().gists().delete("00000000000000000000000000000000").await?;
+    /// # Ok(())
+    /// # }
+    /// ```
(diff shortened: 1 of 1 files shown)


src/api/gists.rs
```diff
@@ -79,6 +79,23 @@ impl<'octo> GistsHandler<'octo> {
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn delete(&self, gist_id: impl AsRef<str>) -> Result<bool> {
+        let gist_id = gist_id.as_ref();
+        let response = self
+            .crab
+            ._delete(format!("/gists/{gist_id}"), None::<&()>)
+            .await?;
+        Ok(response.status() == http::StatusCode::NO_CONTENT)
```
envp, line 96: https://docs.github.com/en/rest/gists/gists?apiVersion=2022-11-28#delete-a-gist--status-codes

funnily enough, it returns a 404 on error, does that get transformed into the `Error(...)` variant here? If so the correct return type might be `Result<()>` since that is isomorphic to `bool`.

Noticed the same _issue_ with status codes from `GET /gists/{gist_id}/star`. WDYT? @XAMPPRocky

XAMPPRocky: Yes, `Result<()>` should be the type here, I think that makes the most sense.

XAMPPRocky:
Thank you for your PR!

PR #343, opened 2023-04-17
ci: cargo fmt

maflcko:
It would be nice to always have a correctly formatted source code. Otherwise it is frustrating to use `cargo fmt` locally during development, when unrelated source code is formatted.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,6 +21,5 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - run: cargo fmt
-    - run: git diff --exit-code
+    - run: cargo fmt --all -- --check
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -21,4 +21,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - run: cargo fmt
+    - run: git diff --exit-code
```
envp, line 25:
```suggestion
    - run: cargo fmt --all -- --check
```

Should be enough to print the diff & exit with a non-zero code.

envp (changes requested):

envp (approved):
Thanks for the change! Definitely saves running `git checkout -p` a bunch of times :D

Issue #346, opened 2023-04-18
hyper-rustls isn't actually supported (v0.20.0)


jsurany:
## error

When building without `tls` (`native-tls`) and enabling `hyper-rustls` in my configuration

```toml
# Cargo.toml / dependencies
octocrab = { version = "0.20", default-features = false, features = ["hyper-rustls", "retry", "stream", "timeout", "tracing"] }
```

attempting to connect gives me the following error

```sh
Error: Service Error: error trying to connect: invalid URL, scheme is not http
```

<details><summary>trace logs</summary>

[2023-04-18T17:27:03Z TRACE tower::buffer::service] sending request to buffer worker
[2023-04-18T17:27:03Z TRACE tower::buffer::worker] worker polling for next message
[2023-04-18T17:27:03Z TRACE tower::buffer::worker] processing new request
[2023-04-18T17:27:03Z TRACE tower::buffer::worker] resumed=false message=worker received request; waiting for service readiness
[2023-04-18T17:27:03Z DEBUG tower::buffer::worker] service.ready=true message=processing request
[2023-04-18T17:27:03Z DEBUG octocrab] HTTP; http.method=GET http.url=https://api.github.com/repos/REDACTED/files otel.name="HTTP" otel.kind="client"
[2023-04-18T17:27:03Z TRACE tracing::span::active] -> HTTP;
[2023-04-18T17:27:03Z DEBUG octocrab] requesting
[2023-04-18T17:27:03Z TRACE tracing::span::active] <- HTTP;
[2023-04-18T17:27:03Z TRACE tower::buffer::worker] returning response future
[2023-04-18T17:27:03Z TRACE tower::buffer::worker] worker polling for next message
[2023-04-18T17:27:03Z TRACE tracing::span::active] -> HTTP;
[2023-04-18T17:27:03Z TRACE hyper::client::pool] checkout waiting for idle connection: ("https", api.github.com)
[2023-04-18T17:27:03Z TRACE hyper::client::connect::http] Http::connect; scheme=Some("https"), host=Some("api.github.com"), port=None
[2023-04-18T17:27:03Z TRACE hyper::client::pool] checkout dropped for ("https", api.github.com)
[2023-04-18T17:27:03Z TRACE hyper::client::pool] checkout waiting for idle connection: ("https", api.github.com)
[2023-04-18T17:27:03Z TRACE hyper::client::connect::http] Http::connect; scheme=Some("https"), host=Some("api.github.com"), port=None
[...]

PR #351, opened 2023-04-23
Enable rustls(and use as default client)

L1ghtman2k:
Enable rustls as per #346

Since rustls is enabled, I have also renamed `tls` feature to `opentls`.

I have implemented the client using `https_or_http`, instead of `https_only`, since the mock server we use for tests doesn't currently support https endpoints. I could alternatively implement two separate clients using conditional compilation for tests, though it becomes rather ugly, especially having to introduce a separate feature just to compile this in integration tests(since `test` flag doesn't propagate to library code)


XAMPPRocky:
Thank you for your PR! I think `https_or_http` is fine for now.

PR #354, opened 2023-05-03
Fix authentication endpoints

rsdy:
After upgrading from a previous `octocrab` version, I've run into an URI parsing regression on these code paths.

I'm not entirely sure if this patch is exhaustive of this pattern/misuse being present in octocrab, but it fixes my issue.

If you'd like me to find a more systematic approach to this, I'm happy to patch that in too.


XAMPPRocky:

PR #359, opened 2023-05-13
Handle redirects for download_tarball

khuey:
I've observed that the GH API endpoint for download_tarball will return a 302. I suspect this was regressed when reqwest was removed because reqwest automatically follows redirects by default. There may be other API endpoints that are currently broken because of the underlying issue ...


XAMPPRocky:
Thank you for your PR! It looks like you need to fix up some fmt or clippy fixes

khuey:
alright looks good now

XAMPPRocky:
Thank you for your PR!

PR #371, opened 2023-05-20
Gists API: Complete support

### LINKED ISSUE #27: Add Gists API

envp:
This PR completes support for GitHub's gists api.

The remaining 3 endpoints in https://github.com/XAMPPRocky/octocrab/issues/27#issuecomment-1500956065 will all be implemented here:

- [x] List gists for a user (GET /users/{username}/gists)
- [x] List all public gists (GET /gists)
- [x] List public gists (GET /gists/public)

Closes #27


src/api/gists/list_gists.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// Sends the actual request.
+    pub async fn send(self) -> crate::Result<crate::Page<crate::models::gists::Gist>> {
+        self.crab.get("/gists", Some(&self)).await
+    }
+}
+
+/// Handles query data for the `GET /gists/public` endpoint.
+#[derive(Debug, serde::Serialize)]
+pub struct ListAllPublicGistsBuilder<'octo> {
```
envp, line 59: There is significant overlap between `ListAllPublicGistsBuilder` and `ListAllGistsBuilder`.

An alternative implementation I considered was to have a flag in `ListAllGistsBuilder` decide which endpoint to call, but decided against it because `GET /gists/public` offers specific guarantees on the order of results returned.

I've no strong opinions on the structure otherwise, happy to merge these two types if that is considered more desirable.

src/api/users.rs
```diff
@@ -0,0 +1,105 @@
...
+    ///         .send()
+    ///         .await?;
+    /// # }
+    /// ```
+    ///
+    /// [docs]: https://docs.github.com/en/rest/gists/gists?apiVersion=2022-11-28#list-gists-for-a-user
+    pub fn list_user_gists(&self, username: impl AsRef<str>) -> ListGistsForUserBuilder<'octo> {
+        ListGistsForUserBuilder::new(self.crab, username.as_ref().to_string())
+    }
+}
```
envp, line 49: Official docs file this endpoint with the rest of gist endpoints. Would you prefer I move this handler into `api/gists.rs` as well?

src/api/gists/list_gists.rs
XAMPPRocky: I think separate structs is probably best as it future proofs us from GitHub making them slightly different in the future.

Another approach you could have taken is to have a type parameter for the visibility, that way it would technically be two different types `ListGistsBuilder<All>` and `ListGistsBuilder<Public>` but they could share all of the same fields.

src/api/users.rs
XAMPPRocky: Sounds good to me.

src/api/gists/list_gists.rs
envp: Interesting, will look into this!

XAMPPRocky:
Thank you for your PR!

PR #373, opened 2023-05-20
Add "updated since" support to ListIssuesBuilder

tgecho:
Allows only returning issues updated since a timestamp.

Docs: https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#list-issues-assigned-to-the-authenticated-user

> since string
>
> Only show notifications updated after the given time. This is a timestamp in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format: YYYY-MM-DDTHH:MM:SSZ.


XAMPPRocky:

Issue #376, opened 2023-05-22
Missing Github Installation API methods


lswith:
https://docs.github.com/en/rest/apps/installations?apiVersion=2022-11-28#list-repositories-accessible-to-the-app-installation

The currently missing API methods are:
- [ ] GET `/installation/repositories`
- [ ] DELETE `/installation/token`
- [ ] GET `/user/installations`
- [ ] GET `/user/installations/<installation id>/repositories`
- [ ] PUT `/user/installations/1/repositories/REPOSITORY_ID`
- [ ] DELETE `/user/installations/1/repositories/REPOSITORY_ID`

Happy to implement them, but I'm not sure where these api methods should go.

XAMPPRocky:
> Happy to implement them, but I'm not sure where these api methods should go.

Thank you for your issue! Generally you can look at how `octokit.js/rest` presents its API, as we generally want a similar API so it feels familiar to people who've used it. From looking at the docs a lot of these methods should be added to `Octocrab::current` since they require an authorisation token to be valid.

https://docs.rs/octocrab/latest/octocrab/current/struct.CurrentAuthHandler.html

PR #378, opened 2023-05-22
Add method for listing a users installations

lswith:
Implements part of #376


XAMPPRocky:
Thank you for your PR!

PR #379, opened 2023-05-24
Add allow_forking & allow_update_branch in Repository model

ledoyen:
It's well hidden in the response schema but it's here :)
https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository

These setting reflects toggles of the same name in the General settings page.
image
image

JSON schema extract:
```json
"allow_update_branch": {
    "type": "boolean",
    "examples": [
        true
    ]
},
"allow_forking": {
    "type": "boolean",
    "examples": [
        true
    ]
},
```

Also present in paths `parent`, `source` and `template_repository`.


XAMPPRocky:

PR #386, opened 2023-05-31
add some of the missing fields to PullRequest

0xB10C:
This adds some of the fields that are missing from the PullRequest model but are present in the GitHub API response. However, I haven't checked if all fields from the API are present in the model now.

The docs can be found here https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request

### DIFF
--- a/src/models/pulls.rs
+++ b/src/models/pulls.rs
@@ -83,6 +83,18 @@ pub struct PullRequest {
     pub draft: Option<bool>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub repo: Option<Box<Repository>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub additions: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub deletions: Option<u64>,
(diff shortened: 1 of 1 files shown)


0xB10C:
Based on my recent experience using the GitHub API, it's not always given that even if the field should always be set, it is always set (see e.g. #388). Though I agree that it's likely not needed here. I'll do a test-run without Options on my use-case and see if there are any `null`s returned.

XAMPPRocky:
Thank you for your PR! Is there a reason you marked the PR as a draft?

0xB10C:
> Is there a reason you marked the PR as a draft?

Wanted to verify if the Option around the fields is needed before this is merged.

Turns out, it's needed. The does not return e.g. `additions` when listing pull requests with https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests but does return `additions` when getting a single pull request with https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#get-a-pull-request.

XAMPPRocky:
Thank you for your PR!

PR #389, opened 2023-06-01
Add Issue Timeline API

0xB10C:
This adds the [Timeline Events API](https://docs.github.com/en/rest/issues/timeline?apiVersion=2022-11-28) for issues along with previously missing [Issue Event Types](https://docs.github.com/en/webhooks-and-events/events/issue-event-types). A `ListTimelineEventsBuilder` is implemented for `Issue` (though as "each pull is also an issue" this works for requesting a Pull-Request timeline too).

I've extensively tested this against the issues and pulls in the bitcoin/bitcoin repo. This resulted in nearly every `TimelineEvent` field besides the actual `event` needing to be `Option<>`s as the API sometimes returns null values for fields it normally sets. Also, some of the added event types aren't listed in the API documentation. I hope I got all of them, but there might be some that don't appear in bitcoin/bitcoin.


XAMPPRocky:
Thank you for your PR!

src/models/timelines.rs
```diff
@@ -0,0 +1,130 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub submitted_at: Option<chrono::DateTime<chrono::Utc>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub state: Option<pulls::ReviewState>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub dismissed_review: Option<DismissedReview>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub pull_request_url: Option<Url>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub requested_reviewers: Option<Vec<Author>>,
```
0xB10C, line 81: This should have been `pub requested_reviewers: Option<Author>`. It's only a single reviewer that's requested per event and not multiple.

0xB10C: fixed with https://github.com/XAMPPRocky/octocrab/pull/390

PR #403, opened 2023-07-08
Add Get Commit API

cloud303-cholden:
Add `get` method to `CommitHandler`. This appears to more reliably return `Some` for optional response fields such as `files` and `stats`. See https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit


XAMPPRocky:
Seems like you just need to run `cargo fmt` (also check `cargo clippy` just in case).

cloud303-cholden:
@XAMPPRocky Sorry should have checked out your workflows before creating the PR. Should be good now.

XAMPPRocky:

PR #408, opened 2023-07-11
Add installation_id in Webhook events

### LINKED ISSUE #407: Add `installation_id: Option<InstallationId>` to models::events::Event
Hello,

I'm currently trying to write a wrapper library to write github applications that react to webhooks. One thing I noticed is that I can't currently access the installation ID from webhook payloads, although the API documentation says that `installation` information is [common in all webhook payloads](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-common-properties), just like the other fields currently living at the "top-level" of [`models::events::Event` structure](https://github.com/XAMPPRocky/octocrab/blob/f2ce7986fc12c7084ff84bf3e022aa53b8dfa39f/src/models/events.rs#L15-L27).

Would it be possible to add access to that id property? From the [current integration tests](https://github.com/XAMPPRocky/octocrab/blob/f2ce7986fc12c7084ff84bf3e022aa53b8dfa39f/tests/resources/workflow_run_event.json#L340), it looks like only the `id` property would be useful ("don't know, don't care" about the `node_id`); and having a `pub installation_id: Option<InstallationId>` in `Event` would be really really nice!

Thanks for the lib anyways,
Gerry

gagbo:
This PR does 2 things:

- Add `pretty_assertions` as a dev dependency to have better output from the "there and back again" serde tests
- Add support for reading `installation_id` from webhook payloads.

This last change is a breaking change (hence the `0.26` bump), because it wraps the `EventPayload` enum to have
easy access to common Event fields whichever variant is actually deserialized.

Tests have been changed accordingly to follow the new API, which could definitely be changed. All the name of the types/fields are also "quick names" that I chose to get going, I'm open to suggestions for better names

Fixes #407

- [ ] Extract sender, organization, and repository from https://github.com/XAMPPRocky/octocrab/blob/f2ce7986fc12c7084ff84bf3e022aa53b8dfa39f/src/models/events/payload/workflow_run.rs#L15 to the common fields of payload (as per [the doc](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-common-properties))

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ tokio-test = "0.4.2"
 wiremock = "0.5.3"
 crypto_box = { version = "0.8.2", features = ["seal"] }
 base64 = "0.21.2"
+pretty_assertions = "1.4.0"
 
 [features]
 default = ["rustls", "timeout", "tracing", "retry"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "octocrab"
-version = "0.25.1"
+version = "0.26.0"
 authors = ["XAMPPRocky <xampprocky@gmail.com>"]
 edition = "2018"
 readme = "README.md"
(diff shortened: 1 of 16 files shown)


XAMPPRocky:

PR #424, opened 2023-07-21
Fix clippy 1.71 warnings

MarcoFalke:
A few changes:

* Run clippy on `--examples` in CI as well.
* Remove `content_type: ContentType` from the search API. This seems unrelated, according to https://docs.github.com/en/search?query=content_type it isn't used in the search API. There is only a TextMatch in the graphql API https://docs.github.com/en/graphql/reference/objects#textmatch, but this is also unrelated because this is the REST API.
* Fix https://rust-lang.github.io/rust-clippy/master/index.html#/non_minimal_cfg warnings

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -22,4 +22,4 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - run: cargo fmt --all -- --check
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR!

PR #427, opened 2023-07-27
Add Webhook events payload deserialization

### LINKED ISSUE #419: missing fields in some structs?
I'm using `octocrab`'s event models to parse webhook events and i've noticed some fields are missing. What's the best way to deal with this? Is there a way to access the rest of the fields? or maybe just make a PR adding the relevant fields?

gagbo:
This commit will be force pushed regularly, as progressively elements
for which we have example payloads will be more precise than just
`serde_json::Value`. This commit is pushed to enable an initial review
over the architecture/scope of the changes.

Closes #419


gagbo:
The main issue with the PR is that there is not a lot of documentation about webhook events schema. So I will only fill the data for the payloads I got examples for, and mostly leave "optimistic" types where possible.

XAMPPRocky:
Thank you for your PR! Feel free to make a follow-up PR with the documentation. 🙂

PR #430, opened 2023-07-31
Make models::webhook_events::payload public

gagbo:
While updating my WIP github app to use the new types, I noticed that I forgot to make the webhook specific payload variants public...

Sorry :(

### DIFF
--- a/src/models/webhook_events.rs
+++ b/src/models/webhook_events.rs
@@ -65,7 +65,7 @@
 //! but the API needs to change according to actual usage in order to refine the different payloads
 //! received.
 
-mod payload;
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "octocrab"
-version = "0.29.0"
+version = "0.29.1"
```
XAMPPRocky, line 3: Can you remove the version upgrade? That's automatically handled by `release-plz`.

XAMPPRocky:
Thank you for your PR!

PR #433, opened 2023-08-01
Add `get_org_installation` to `AppsRequestHandler`

marcusirgens:
Similarly to `get_repository_installation`, this fetches the installation for the given org.

Sample response payload copied from the [API docs](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#get-an-organization-installation-for-the-authenticated-app).

### DIFF
--- a/src/api/apps.rs
+++ b/src/api/apps.rs
@@ -97,4 +97,14 @@ impl<'octo> AppsRequestHandler<'octo> {
 
         self.crab.get(&route, None::<&()>).await
     }
+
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR! Seems like you just need to format the code 🙂

marcusirgens:
> (...) Seems like you just need to format the code 🙂

Yep, and fix some invalid comments after over-eager code copying 😅

XAMPPRocky:

PR #440, opened 2023-08-15
add `WebhookEventType::Schedule` variant

JeanMertz:
I added this because I'm using the `WebhookEventType` to strongly type the event data available during a GitHub Workflow run. However, it's not strictly a webhook event, as it can't be received as a webhook.

You can read about this event on the "[Events that trigger workflows](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule)" page, but it's (for obvious reasons) omitted from the "[Webhook events and payloads](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads)" page.

Most (all except this one?) events listed on the first page, link to the relevant webhook events on the second page, so I don't think there's a need to have a completely separate type, but if you feel differently, I'd be happy to make a change.


XAMPPRocky:

PR #436, opened 2023-08-04
Allow storing, extracting installation tokens

marcusirgens:
First attempt to handle #435.

Most of the changes in this pull request are from the added test.

I made some breaking changes, which I am happy to discuss or to solve differently. The reasoning behind this is that adding a method on `octocrab::Octocrab` caused a lot of stutter:

```rust
impl Octocrab {
    pub fn installation(&self, id: InstallationId) -> Self { /* ... */ }
    pub async fn installation_and_token(&self, id: InstallationId) -> Result<(Self, InstallationToken)> { /* ... */ }

    // new?
    pub fn installation_with_token(&self, id: InstallationId, token: InstallationToken) -> Self { /* ... */ }
    pub async fn current_installation_token(&self, force_refetch: bool) -> Result<InstallationToken> { /* ... */ }
}
```

The approach I suggest gives the following API:

```rust
impl Octocrab {
    pub fn installation(&self, id: InstallationId) -> InstallationClientBuilder { /* ... */ }
    pub async fn installation_token(&self, force_refetch: bool) -> Result<InstallationToken> { /* ... */ }
}
```

## `InstallationClientBuilder`
`octocrab::Octocrab::installation()` now emits an `octocrab::InstallationClientBuilder`.

### Building without a known token (this is probably the most significant breaking change):

Prior to this change:
```rust
let installation_client: octocrab::Octocrab = app_client.installation(123.into());
```

After this change:
```rust
let installation_client: octocrab::Octocrab = app_client.installation(123.into()).build();
```

### Building with a known token

Not possible prior to this change.

After this change:
```rust
let installation_client: octocrab::Octocrab = app_client
    .installation(123.into())
    .with_token(my_token)
    .build();
```

## Building a client and emitting the token

Prior to this change:
```rust
let (installation_client, token): (octocrab::Octocrab, octocrab::models::InstallationToken) = app_client.installation_and_token(123.into).await?;
```

After this change:
```rust
let installation_client: octocrab::Octocrab = app_client.installation(123.into()).build();
let token = installation_client.installation_token(/* force_refetch: */ false).await?;
```

## Fetching the token from the client

The token can now be extracted. Here's the gist of how we're going to use this internally, which allows us to significantly reduce how many requests we send to GitHub.

```rust
async fn our_business_logic(app_client: octocrab::Octocrab, token_cache: TokenCache, customer: Customer) -> Result<(), Error> {
    let installation_client = match token_cache.get_installation(customer.id).await? {
[...]

PR #442, opened 2023-08-16
Fix installation token cache issue

matthewgapp:
My application holds an Installation for an extended period of time (longer than 1 hour). This causes a "Bad credentials" error from Github. This PR introduces a fix that changes the installation token cache to only return a token if it's not expired, forcing Octocrab client to refresh the token appropriately.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -791,14 +791,26 @@ impl CachedToken {
     }
 
     fn get_valid_token_with_buffer(&self, buffer: chrono::Duration) -> Option<SecretString> {
-        if let Some(expiration) = self.0.read().unwrap().as_ref().and_then(|s| s.expiration) {
-            if expiration - Utc::now() > buffer {
-                self.get_token()
-            } else {
-                None
-            }
-        } else {
+        let inner = self.0.read().unwrap();
+
+        // If there's no token reference, return immediately.
+        let token = match inner.as_ref() {
+            Some(token) => token,
+            None => return self.get_token(),
+        };
+
+        let expiration = match token.expiration {
+            Some(expiration) => expiration,
+            // return early with the token if there's no expiration
+            None => return self.get_token(),
+        };
+
+        let is_expiring = expiration - Utc::now() <= buffer;
+
+        if is_expiring {
             None
+        } else {
+            self.get_token()
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -760,18 +761,54 @@ impl DefaultOctocrabBuilderConfig {
...
+            } else {
+                return Some(token.secret.clone());
+            }
+        }
+
+        None
+    }
+
+    fn valid_token(&self) -> Option<SecretString> {
+        self.valid_token_with_buffer(chrono::Duration::seconds(30))
```
matthewgapp, line 806: Consider a token that expires within 30 seconds to be expired by default. 

From what I can tell, the expiration is usually set about 1 hour from token mint.

src/lib.rs
```diff
@@ -1349,7 +1386,21 @@ impl Octocrab {
 
         let token_object =
             InstallationToken::from_response(crate::map_github_error(response).await?).await?;
-        token.set(token_object.token.clone());
+
+        let expiration = token_object
+            .expires_at
+            .map(|time| {
+                DateTime::<Utc>::from_str(&time).map_err(|e| error::Error::Other {
```
matthewgapp, line 1393: Github doesn't document the timezone but I assumed it was UTC. Also checked a couple responses manually and verified that it was UTC.

XAMPPRocky:
Thank you for your PR! However I need to check with someone to make sure these aren't overlapping (and different) proposals for the same feature.

cc @marcusirgens Is this problem solved by #436 or is it separate?

marcusirgens:
@XAMPPRocky Yes, I solve this very similarily in #436, making `CachedToken::get` return `None` if the token is expired.

I'm not opposed to merging this first and working it into my PR, as it's more focused than my "do many things at once"-request, especially considering that #436 needs some clarifications. :)

https://github.com/marcusirgens/octocrab/blob/store-installation-tokens/src/lib.rs#L811-L828

matthewgapp:
> I'm not opposed to merging this first and working it into my PR, as it's more focused than my "do many things at once"-request, especially considering that #436 needs some clarifications. :)

@XAMPPRocky sounds like it makes sense to merge so that @marcusirgens can rebase onto this.

XAMPPRocky:
Thank you for your PR!

PR #450, opened 2023-08-30
Add CommitHandler::associated_check_runs

iamjpotts:
The crate currently supports getting the check runs of a check suite.

This adds support for getting the check runs of a specific commit, such as the last commit of a branch related to a pull request, or the last commit in the default branch.

PR #451, opened 2023-08-30
Add Octocrab::users() and UsersHandler::repos

iamjpotts:
Add a new `Octocrab::users` api with a single method to get the list of repositories associated with a user, `UsersHandler::repos`.

I am not sure if the documentation link I added to the README will be valid, or if additional steps are required for that change. It was added on the assumption the html file would be auto-generated by a build or publish process.


iamjpotts:
Removed CHANGELOG entry after seeing comment on https://github.com/XAMPPRocky/octocrab/pull/450#issuecomment-1700536869.

XAMPPRocky:
Thank you for your PR!

Issue #452, opened 2023-09-01
commit_comment.created events are not deserialized properly in webhook events


gagbo:
Similar to other deserialization errors, I found a new payload that doesn't work with current webhook events.

## Payload

Header : `X-GitHub-Event` is `commit_comment`

```json
{
    "action": "created",
    "comment": {
        "url": "https://api.github.com/repos/gagbo/app-test-repo/comments/126171404",
        "html_url": "https://github.com/gagbo/app-test-repo/commit/5a722779076a943e9e8ccc502566c411f6481b4a#commitcomment-126171404",
        "id": 126171404,
        "node_id": "CC_kwDOKIoqtM4HhTkM",
        "user": {
            "login": "gagbo",
            "id": 10496163,
            "node_id": "MDQ6VXNlcjEwNDk2MTYz",
            "avatar_url": "https://avatars.githubusercontent.com/u/10496163?v=4",
            "gravatar_id": "",
            "url": "https://api.github.com/users/gagbo",
            "html_url": "https://github.com/gagbo",
            "followers_url": "https://api.github.com/users/gagbo/followers",
            "following_url": "https://api.github.com/users/gagbo/following{/other_user}",
            "gists_url": "https://api.github.com/users/gagbo/gists{/gist_id}",
            "starred_url": "https://api.github.com/users/gagbo/starred{/owner}{/repo}",
            "subscriptions_url": "https://api.github.com/users/gagbo/subscriptions",
            "organizations_url": "https://api.github.com/users/gagbo/orgs",
            "repos_url": "https://api.github.com/users/gagbo/repos",
            "events_url": "https://api.github.com/users/gagbo/events{/privacy}",
            "received_events_url": "https://api.github.com/users/gagbo/received_events",
            "type": "User",
            "site_admin": false
        },
        "position": null,
        "line": null,
        "path": null,
        "commit_id": "5a722779076a943e9e8ccc502566c411f6481b4a",
        "created_at": "2023-09-01T12:01:16Z",
        "updated_at": "2023-09-01T12:01:16Z",
        "author_association": "OWNER",
        "body": "@gagbo-test-app[bot] compare-tag v0.1",
        "reactions": {
            "url": "https://api.github.com/repos/gagbo/app-test-repo/comments/126171404/reactions",
            "total_count": 0,
            "+1": 0,
            "-1": 0,
            "laugh": 0,
            "hooray": 0,
            "confused": 0,
            "heart": 0,
            "rocket": 0,
[...]

PR #453, opened 2023-09-01
Fix commit_comment webhook event parsing

gagbo:
The list of reactions in commit_comment events was badly represented.
The correct type has directly been added in the models::commits::Comment
struct, so everything has the better typing out of the box.

This also enforces strong typing with the author_association field.


gagbo:
Fixes #452

XAMPPRocky:
Thank you for your PR!

PR #458, opened 2023-09-06
Fix `github_app_authentication.rs` example

### LINKED ISSUE #215: Github App authentication example fails
Hello,

first of all thanks for all your effort resulting in this wonderful library :)

I'm confused as to how to authenticate as a Github App, and I'm unable to reproduce the example.

Steps I took:

* [Register a Github App](https://github.com/settings/apps)
* make a note of the *App Id*
* Generate a private key for it at https://github.com/settings/apps/<bot-name> (bottom of the page)
* From inside the octocrab repo:
  ```sh
    export GITHUB_APP_id=123456
    export GITHUB_APP_PRIVATE_KEY=$(cat <bot-name>.<date>.private-key.pem)
    cargo run --example github_app_authentication
  ```

This will result in a (401, "Bad credentials") response from Github.

Am I misunderstanding something here, or did I run into an issue?

Thank you for your help :)

dima74:
Fixes #215

The API call `.get("/installation/repositories", ...)`  has to be authenticated through an installation ID, so better to use `.apps().installations()` API call in this example

### DIFF
--- a/examples/github_app_authentication.rs
+++ b/examples/github_app_authentication.rs
@@ -8,11 +8,7 @@ async fn main() -> octocrab::Result<()> {
     let key = jsonwebtoken::EncodingKey::from_rsa_pem(app_private_key.as_bytes()).unwrap();
 
     let octocrab = Octocrab::builder().app(app_id, key).build()?;
-    let installed_repos: InstallationRepositories = octocrab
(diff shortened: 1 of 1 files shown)


XAMPPRocky:

PR #460, opened 2023-09-12
Make `IssueHandler::delete_label` return `()`

### LINKED ISSUE #456: `IssueHandle::delete_label` panics
(Code is using `octocrab = "0.30.1"`)
When calling `IssueHandle::delete_label`, `IssueHandle` panics. This appears to be because it expects a response in Json format, but according to [github's documentation](https://docs.github.com/en/rest/issues/labels?apiVersion=2022-11-28#delete-a-label), the endpoint does not return any data (on success, it yields 204 No Content).

chorman0773:
Note: This may require a major version increase.
Fixes #456

### DIFF
--- a/src/api/issues.rs
+++ b/src/api/issues.rs
@@ -472,7 +472,7 @@ impl<'octo> IssueHandler<'octo> {
     /// # Ok(())
     /// # }
     /// ```
-    pub async fn delete_label(&self, name: impl AsRef<str>) -> Result<models::Label> {
+    pub async fn delete_label(&self, name: impl AsRef<str>) -> Result<()> {
         let route = format!(
             "/repos/{owner}/{repo}/labels/{name}",
             owner = self.owner,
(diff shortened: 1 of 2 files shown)


src/api/issues.rs
```diff
@@ -472,15 +472,17 @@ impl<'octo> IssueHandler<'octo> {
...
+    pub async fn delete_label(&self, name: impl AsRef<str>) -> Result<()> {
         let route = format!(
             "/repos/{owner}/{repo}/labels/{name}",
             owner = self.owner,
             repo = self.repo,
             name = name.as_ref(),
         );
 
-        self.crab.delete(route, None::<&()>).await
+        let _: crate::from_response::NoBody = self.crab.delete(route, None::<&()>).await?;
```
XAMPPRocky, line 483: You don't need that, you can just write the following

```rust
crate::map_github_error(self.crab._delete(uri, None::<&()>).await?)
    .await
    .map(drop)
```

XAMPPRocky:
@chorman0773 Friendly ping 🙂

manchicken:
It looks like this was fixed by #503.

XAMPPRocky:
Thank you for your PR!

Issue #463, opened 2023-10-02
[Bug] octocrab::Octocrab::request_installation_auth_token fails with status code 411


phideg:
I'm writing a GitHub app that fails at following code snippet:

```rust
    // read issues
    let issue_handler = octocrab.issues(GITHUB_OWNER, GITHUB_REPO);
    let issues = issue_handler
        .list()
        .state(params::State::Open)
        .per_page(10)
        .send()
        .await?
        .take_items();
```

call stack of error is:

```text
Error: Serde Error: expected value at line 2 column 1
Found at  0 <snafu::backtrace_shim::Backtrace as snafu::GenerateImplicitData>::generate::hf12a34d663f0e574
   /home/ph/.cargo/registry/src/index.crates.io-6f17d22bba15001f/snafu-0.7.5/src/backtrace_shim.rs:15
   snafu::GenerateImplicitData::generate_with_source::h647f062bf6690964
   /home/ph/.cargo/registry/src/index.crates.io-6f17d22bba15001f/snafu-0.7.5/src/lib.rs:1154
 1 <octocrab::error::SerdeSnafu as snafu::IntoError<octocrab::error::Error>>::into_error::h5aadde84b6442c0d
   /home/ph/.cargo/registry/src/index.crates.io-6f17d22bba15001f/octocrab-0.31.0/src/error.rs:23
 2 <core::result::Result<T,E> as snafu::ResultExt<T,E>>::context::hb74aa06b1cbf6eec
   /home/ph/.cargo/registry/src/index.crates.io-6f17d22bba15001f/snafu-0.7.5/src/lib.rs:699
 3 octocrab::map_github_error::{{closure}}::h390244a7a3c433b0
   /home/ph/.cargo/registry/src/index.crates.io-6f17d22bba15001f/octocrab-0.31.0/src/lib.rs:306
 4 octocrab::Octocrab::request_installation_auth_token::{{closure}}::h21aa363dc82eab90
   /home/ph/.cargo/registry/src/index.crates.io-6f17d22bba15001f/octocrab-0.31.0/src/lib.rs:1402
 5 octocrab::Octocrab::execute::{{closure}}::haff4e60098dbb58b
   /home/ph/.cargo/registry/src/index.crates.io-6f17d22bba15001f/octocrab-0.31.0/src/lib.rs:1479
 6 octocrab::Octocrab::_get_with_headers::{{closure}}::h1049d18787c232f6
[...]

PR #464, opened 2023-10-04
Always set http request header `CONTENT-LENGTH`

phideg:
This PR fixes issue #463.

The GitHub REST API endpoints, that use the http `POST` method, seem to require a `CONTENT-LENGTH` header even if the request body is empty.

This change addresses the issue by either sending an empty JSON document or by explicitly setting the `CONTENT-LENGTH` header as described in following issue: https://github.com/seanmonstar/reqwest/issues/838


XAMPPRocky:

PR #467, opened 2023-10-09
Add conclusion, completed_at and output fields to create check api

ikkerens:
The [create checks api](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#create-a-check-run) also allows for immediate reporting of its completion status for quick status reports (for example, verifying that a PR name matches a regex).

This PR adds those fields to the create check api and builder, skipping the need for another request to update the check immediately after creating it.

### DIFF
--- a/examples/github_app_authentication.rs
+++ b/examples/github_app_authentication.rs
@@ -1,4 +1,3 @@
-use octocrab::models::InstallationRepositories;
 use octocrab::Octocrab;
 
 #[tokio::main]
--- a/src/api/checks.rs
+++ b/src/api/checks.rs
@@ -20,8 +20,6 @@ pub enum CheckRunStatus {
     Completed,
 }
 
-/// Can be one of `success`, `failure`, `neutral`, `cancelled`, `timed_out`,
-/// `skipped`, `stale` or `action_required`.
 #[derive(serde::Serialize)]
 #[serde(rename_all = "snake_case")]
 pub enum CheckRunConclusion {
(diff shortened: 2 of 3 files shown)


src/api/checks.rs
```diff
@@ -66,6 +75,28 @@ impl<'octo, 'r> CreateCheckRunBuilder<'octo, 'r> {
         self
     }
 
+    /// The final conclusion of the check.
+    /// Can be one of `success`, `failure`, `neutral`, `cancelled`, `timed_out`,
+    /// `skipped`, `stale` or `action_required`.
+    pub fn conclusion(mut self, conclusion: impl Into<String>) -> Self {
```
XAMPPRocky, line 81: Can you make this an enum with a `Display` implementation? That would make it so you can't provide an incorrect variant.

ikkerens: I added the enum type, and made a few other things strongly typed as well.
The Display implementation puzzled me a bit, since most enums don't seem to have this. (Though some have a Debug derive)
Though if you still want me to add them for the enum I've created feel free to let me know.

examples/github_app_authentication.rs
```diff
@@ -1,4 +1,3 @@
-use octocrab::models::InstallationRepositories;
```
ikkerens, line 1: When running tests I got this warning, figured I might as well remove it.

src/api/checks.rs
```diff
@@ -20,6 +20,66 @@ pub enum CheckRunStatus {
     Completed,
 }
 
+/// Can be one of `success`, `failure`, `neutral`, `cancelled`, `timed_out`,
+/// `skipped`, `stale` or `action_required`.
+#[derive(serde::Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum CheckRunConclusion {
```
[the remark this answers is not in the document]
ikkerens: Done!

src/api/checks.rs
XAMPPRocky: Serialising it is also fine, Display is just quicker and faster, because you know you're converting to a string.

ikkerens:
I've ran the formatter to fix some of the issues, though my clippy locally still fails due to code that is not mine.
I take it the github action here is running an older version of rust that does not have the checks present in rust clippy 1.73.0?

XAMPPRocky:

PR #477, opened 2023-10-27
add fields for refresh token and expiry timestamps

SebRollen:
GitHub now supports [expiring access tokens](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/refreshing-user-access-tokens).

For these tokens, GitHub sends additional fields in the OAuth response, namely `expires_in`, `refresh_token` and `refresh_token_expires_in`. This pull request optionally adds these fields to the OAuth response struct so that they can be used by applications that are using expiring access tokens.

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -100,6 +100,9 @@ pub struct OAuth {
     pub access_token: SecretString,
     pub token_type: String,
     pub scope: Vec<String>,
+    pub expires_in: Option<usize>,
+    pub refresh_token: Option<SecretString>,
+    pub refresh_token_expires_in: Option<usize>,
 }
 
 /// The wire format of the OAuth struct.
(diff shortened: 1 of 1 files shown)


src/auth.rs
```diff
@@ -108,6 +111,9 @@ struct OAuthWire {
     access_token: String,
     token_type: String,
     scope: String,
+    expires_in: Option<usize>,
+    refresh_token: Option<String>,
+    refresh_token_expires_in: Option<usize>,
```
SebRollen, line 116: The size of `expires_in` and `refresh_token_expires_in` are currently fixed at `28800` and `15811200`, respectively, so we could size down these fields to be `Option<u16>` and `Option<u32>`. I just copied the precedent for the `Claims` struct here

XAMPPRocky:
Thank you for your PR!

PR #483, opened 2023-11-03
Add some actions API methods for working with self-hosted runners

samkearney:
This PR adds some methods to the ActionsHandler to work with GitHub's self-hosted runners API, for managing self-hosted GitHub Actions runners associated with repos or organizations. I didn't have time to cover the full API, so I implemented the methods I need and what I think are likely to be the most popular methods in this area of the API.

Relevant API reference is [here](https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28).

I have tried to follow the styles and conventions that I see in the project, but please let me know if I missed anything. Thanks!

### DIFF
--- a/examples/list_org_self_hosted_runners.rs
+++ b/examples/list_org_self_hosted_runners.rs
@@ -1,0 +1,36 @@
+use octocrab::Octocrab;
+
+#[tokio::main]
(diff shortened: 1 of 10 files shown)


XAMPPRocky:

PR #481, opened 2023-11-01
examples/graphql_issues.rs: Demonstrate variables and paging

Enselic:
This was quite tricky to figure out how to do (for me, anyway), so let's help others by demonstrating how to do it.

Example output:

```console
$ GITHUB_TOKEN=github_pat_123ABC cargo run --example=graphql_issues
Page 1:
https://github.com/XAMPPRocky/octocrab/issues/435 Installation tokens are hard to re-use
https://github.com/XAMPPRocky/octocrab/issues/456 `IssueHandle::delete_label` panics
https://github.com/XAMPPRocky/octocrab/issues/463 [Bug] octocrab::Octocrab::request_installation_auth_token fails with status code 411
https://github.com/XAMPPRocky/octocrab/issues/474 Add Method For Viewing Repository Traffic Stats
https://github.com/XAMPPRocky/octocrab/issues/475 Add API Version configuration to the builder
Page 2:
https://github.com/XAMPPRocky/octocrab/issues/398 Improve Commits API
https://github.com/XAMPPRocky/octocrab/issues/417 Add Projects API
https://github.com/XAMPPRocky/octocrab/issues/421 How to access commit compare endpoint?
https://github.com/XAMPPRocky/octocrab/issues/422 Add `Status`/`Conclusion` enums directly to fields
https://github.com/XAMPPRocky/octocrab/issues/423 Add a `Page::all` method
Page 3:
https://github.com/XAMPPRocky/octocrab/issues/377 Generating Github REST API from Open-API
https://github.com/XAMPPRocky/octocrab/issues/380 Source closed when authenticating with personal token
https://github.com/XAMPPRocky/octocrab/issues/381 The Workflow Run Event Action `in_progress` is Not Supported
https://github.com/XAMPPRocky/octocrab/issues/394 Add method for downloading job logs
https://github.com/XAMPPRocky/octocrab/issues/396 Initializing octocrab takes 300ms on macOS
```

PR #485, opened 2023-11-04
examples/graphql_issues.rs: Use graphql_client to get stronger typing

Enselic:
This results in nicer code. The biggest downside is that we need the GitHub GraphQL schema. We could download it in a build script, but it's only about a megabyte, and changes rarely, so we can probably get away with versioning a copy of it in the repo. It's nice to not require custom build scripts.

Doing this was suggested in https://github.com/XAMPPRocky/octocrab/pull/481. I don't think we should keep the old example code though, because in the end it is nicer to use a typed API, and I don't think it is worth maintaining an example of an untyped API.


XAMPPRocky:
Thank you for your PR!

Issue #497, opened 2023-12-01
Timeouts


johnrichardrinehart:
How can I build an instance of an `Octocrab` client that uses timeouts? I've enabled the feature but all of the timeout values are [hard-coded to `None`](https://github.com/XAMPPRocky/octocrab/blob/551871a5766339f47119ee84686ce2dd4310e6ba/src/lib.rs#L749-L754) without `Builder` methods that allow for changing these values from `None`.

CommanderStorm:
@johnrichardrinehart I think this issue can be closed given that #498 was merged ^^

PR #498, opened 2023-12-01
feat: add methods to builder for changing timeouts

johnrichardrinehart:
Work related to #497 .

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -486,21 +486,18 @@ impl OctocrabBuilder<NoSvc, DefaultOctocrabBuilderConfig, NoAuth, NotLayerReady>
     }
 
     /// Set the connect timeout.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -485,6 +485,27 @@ impl OctocrabBuilder<NoSvc, DefaultOctocrabBuilderConfig, NoAuth, NotLayerReady>
         self
     }
 
+    /// Set the connect timeout.
+    #[cfg(feature = "timeout")]
```
XAMPPRocky, line 489: Does this really need its own feature flag? I think it's probably better to just have it included by default.

XAMPPRocky (commented):
Thank you for your PR! LGTM I have just one minor change request.

XAMPPRocky:

Issue #490, opened 2023-11-16
Missing `list_contributors`


manchicken:
There's no support for listing contributors, only listing collaborators.

This is important because a "collaborator" is a GitHub user who _could_ commit (or who could modify the repository), and a "contributor" is a GitHub user who _has_ committed.

For GitHub Enterprise Organizations, this includes members of any team which has write or admin privs on a repo, and it also includes anybody who is a GHAS security administrator.

https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repository-contributors

PR #500, opened 2023-12-01
added repos/list_contributors

dmgorsky:
resolving #490


dmgorsky:
@XAMPPRocky

src/api/repos/contributors.rs
```diff
@@ -0,0 +1,52 @@
...
+        Self {
+            handler,
+            anon: None,
+            per_page: None,
+            page: None,
+        }
+    }
+
+    /// Set to 1 or true to include anonymous contributors in results.
+    pub fn anon(mut self, include_anon: impl Into<bool>) -> Self {
```
manchicken, line 26: Dumb question: does it matter that these aren't `&mut self`?

manchicken (commented):
I think it would be nice if you added a test as well.

src/api/repos/contributors.rs
dmgorsky: Here we consume `self` and then return it, as a part of a builder pattern.

manchicken: That makes sense, sorry I missed it the first go-round.

manchicken:
This looks great, I can't wait to use it!

dmgorsky:
@XAMPPRocky This workflow requires approval from a maintainer 😇

XAMPPRocky:

PR #501, opened 2023-12-04
models/timelines: make dismissal_message optional

jecluis:
Assuming `dismissal_message` is always a string seems to conflict with some GitHub API responses, which leads to `serde` being unable to deserialize the payload.

Signed-off-by: Joao Eduardo Luis \<joao@1e3ms.io>


XAMPPRocky:

Issue #197, opened 2022-03-29
Failure parsing JSON for release listing response


urma:
While trying to fetch the releases for a specific repository (https://github.com/protocolbuffers/protobuf) I am getting the following:
```
thread 'main' panicked at 'Unable to find any acceptable when scraping GitHub: Serde Error: invalid type: null, expected struct User
(...)
Caused by:
    invalid type: null, expected struct User', src/dependencies/libprotobuf.rs:48:10
```

Simplest code that can reproduce the issue is:
```rust
let octocrab = octocrab::instance();

// Returns the first page of all releases.
let mut page = octocrab
    .issues("protocolbuffers", "protobuf")
    .list()
    .send()
    .await?;
```

I have found that making the `author` attribute in https://github.com/XAMPPRocky/octocrab/blob/c78edcd87fa5edcd5a6d0d0878b2a8d020802c40/src/models/repos.rs#L166-L185 as well as making the `uploader` attribute in https://github.com/XAMPPRocky/octocrab/blob/c78edcd87fa5edcd5a6d0d0878b2a8d020802c40/src/models/repos.rs#L189-L203 optional (`Option<crate::models::User>`) the code works as expected.

I have checked the JSON response for the `releases` endpoint for that particular repository (https://api.github.com/repos/protocolbuffers/protobuf/releases) and all the required attributes in both the `author` and `uploader` objects for all `release` instances seem to be properly populated.

DrProfSgtMrJ:
I'm curious if `pub name` should be optional as well. I currently get a similar error when attempting to  get a release that has no name.

DrProfSgtMrJ:
So I got a similar issue 
`Error: Failed to get release JSON Error in assets[0].uploader: missing field `name` at line 1 column 2898.`
 I see that it's not optional at the moment.
 Wondering if I should also override this struct... 
 @urma

PR #512, opened 2024-01-08
Make user information optional

djrodgerspryor:
Fixes https://github.com/XAMPPRocky/octocrab/issues/197

Despite it being documented as required in the Github API, the `author` field is sometimes absent when listing releases (perhaps in rare cases where a user's profile is deleted or similar).

This can be demonstrated by listing releases for the segmentio/chamber repo: https://api.github.com/repos/segmentio/chamber/releases
Screenshot 2024-01-08 at 2 17 24 pm

Making this field optional allows octocrab users to handle this partial data gracefully rather than making all of the release data inaccessible.

This is obviously a breaking change, so I'd think that it should go into an 0.33.0 release rather than into a patch.

### DIFF
--- a/src/models/repos.rs
+++ b/src/models/repos.rs
@@ -305,7 +305,7 @@ pub struct Release {
     pub prerelease: bool,
     pub created_at: Option<DateTime<Utc>>,
     pub published_at: Option<DateTime<Utc>>,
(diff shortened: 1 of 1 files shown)


XAMPPRocky:

PR #513, opened 2024-01-08
[docs] Improve the update_file docs

beyarkay:
This is a documentation-only commit. It updates the docs for the update_file function to explain what the different parameters do and adds a hyperlink to the GitHub API docs.

I was using `octocrab` but it wasn't clear what the `sha` parameter was, and why I'd need to provide a file SHA if I'm already giving the path of the file I want to update. I did some digging to figure out the reason, so I thought I'd update the docs to explain this for future travellers.

### DIFF
--- a/src/api/repos.rs
+++ b/src/api/repos.rs
@@ -287,6 +287,17 @@ impl<'octo> RepoHandler<'octo> {
         )
     }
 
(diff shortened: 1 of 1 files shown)


XAMPPRocky:

PR #515, opened 2024-01-09
models: Add CheckRunOutput

maflcko:
Per https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#get-a-check-run


XAMPPRocky:
Thank you for your PR!

PR #519, opened 2024-01-15
fixed timeout methods not being feature gated

CommanderStorm:
> [!NOTE]
> For those affected: Add the `timeout`-feature to get around this bug

In #498 a few methods were introduced which reference fields which are feature-gated.

image

causing 
image

This PR applies the same feature gate for these methods (reverted https://github.com/XAMPPRocky/octocrab/pull/498/commits/f1c94c14e250cf7543a8f9bae273613ef5f7ae55)

Cheers and thanks for providing such an usefull crate ^^

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ http = "0.2.5"
 http-body = "0.4.5"
 hyper = { version = "0.14.13", features = ["client", "http1", "stream", "tcp"] }
 hyper-rustls = { version = "0.24.0", optional = true }
-hyper-timeout = { version = "0.4.1" }
+hyper-timeout = { version = "0.4.1", optional = true }
 hyper-tls = { version = "0.5.0", optional = true }
 once_cell = "1.7.2"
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR! However, I think what happened is that we forgot to remove the feature gate everywhere, so I'd prefer to remove all mentions of `#[cfg(feature = "timeout")]`

CommanderStorm:
I have refactored this PR as such.
Note that this folds `tokio` and `hyper-timeout` into the non-disablable parts of the lib. 
Idk if this is intended (I don't know enough about rust to have an opinion on this ^^)

XAMPPRocky:
Ah okay, sorry I guess I didn't understand that this would then require tokio, sorry for the extra work but if you revert to your original PR, we can merge that 🙂

CommanderStorm:
> sorry for the extra work

No worries ^^

Have reverted the commit

XAMPPRocky:

PR #556, opened 2024-01-25
Handle redirects in ReleasesHandler::stream_asset().

korran:
From
https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28#get-a-release-asset

```
API clients should handle both a 200 or 302 response.
```


manchicken:
This looks cool! I recommend adding a test for this case to help prevent regression.

(I am not the maintainer, I'm just another contributor, so my suggestion is just a suggestion.)

XAMPPRocky:
Thank you for your PR!

PR #560, opened 2024-02-04
Added tests for `repos().merges()`

manchicken:
- Added a test for 201 cases
- Added a test for 204 cases
- Fixed the broken 204 case

**BREAKING CHANGE WARNING**: This changes the
return type of the builder from
`Result<models::repos::MergeCommit>` to
`Result<Option<models::repos::MergeCommit>>`.

### DIFF
--- a/src/api/repos/merges.rs
+++ b/src/api/repos/merges.rs
@@ -1,4 +1,5 @@
 use super::*;
+use crate::from_response::FromResponse;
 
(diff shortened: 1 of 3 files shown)


manchicken:
I do think that this is the cleanest approach to this problem, but I don't like introducing breaking changes.

XAMPPRocky:
Thank you for your PR!

PR #562, opened 2024-02-05
Don't send Authorization headers to third-party services.

korran:
We may have been directed away from github.com (via follow_location_to_data()), and for security reasons it's important that we don't leak our GitHub credentials.

If an artifact is uploaded with actions/upload-artifact@v4, calls may be redirected to a URL starting with
https://productionresultssa0.blob.core.windows.net/. Requests made to download that file will be rejected if they contain an Authorization header.


XAMPPRocky:
Thank you for your PR!

PR #570, opened 2024-02-09
Added tests to prove `repos().secrets().delete_secret()`

manchicken:
- Added test for 204 case
- Added test for 500 case


XAMPPRocky:
Thank you for your PR!

PR #578, opened 2024-02-15
Added some more tests

manchicken:
- `repos().issues().check_assignee()`
  - Covers 204 case
  - Covers 404 case
  - Covers 500 case


XAMPPRocky:
Thank you for your PR!

PR #579, opened 2024-02-15
Fixing the test that `wiremock@v0.6.0` breaks.

manchicken:
- Updated wiremock to 0.6.0
- Fixed the one test broken by the update

I did this one in response to seeing tests fail here: https://github.com/XAMPPRocky/octocrab/pull/575


XAMPPRocky:
Thank you for your PR!

PR #585, opened 2024-02-23
Added more tests, for `repos().delete()`

manchicken:
- 204 case
- 404 case
- 500 case


XAMPPRocky:
Thank you for your PR!

PR #587, opened 2024-02-23
Fixed the `projects().delete_project().send() function

manchicken:
- This endpoint, `DELETE /projects/{project_id}` [is documented](https://docs.github.com/en/enterprise-cloud@latest/rest/projects/projects?apiVersion=2022-11-28#delete-a-project) as never returning 200.
- The test was incorrectly verifying a 200 case, and the function broke in a 204
- I fixed the function
- I fixed the test
- I added coverage for 410 and 500 cases, too


XAMPPRocky:
Thank you for your PR!

PR #589, opened 2024-02-27
Add PublicEvent and ReleaseEvent

LuaKT:
Hello,
This adds the `ReleaseEvent` and `PublicEvent` from https://docs.github.com/en/rest/using-the-rest-api/github-event-types. The `PublicEvent` doesn't have a payload so I have used an empty struct, not sure if there's a better way to handle it.

Thanks!


XAMPPRocky:
Thank you for your PR!

Issue #571, opened 2024-02-09
Create commit from tree SHA


tniessen:
I'm trying to migrate from octokit to octocrab and I am using this workaround to create a commit from an existing tree SHA without updating any git references:

```rust
#[derive(Deserialize)]
#[serde(rename_all = "snake_case")]
#[non_exhaustive]
struct CreatedCommit {
    pub sha: String,
}

async fn create_commit(
    octocrab: &octocrab::Octocrab,
    owner: &str,
    repo: &str,
    message: &str,
    tree_sha: &str,
) -> octocrab::Result<CreatedCommit> {
    let route = format!("/repos/{owner}/{repo}/git/commits");

    octocrab
        .post(
            route,
            Some(&serde_json::json!({
                "message": message,
                "tree": tree_sha
            })),
        )
        .await
}
```

However, it would be great if octocrab had a built-in function to do this, including all relevant parameters and result fields, similar to octokit. (Perhaps it already exists and I simply haven't found it yet!)

dmgorsky:
like this?
https://github.com/XAMPPRocky/octocrab/pull/600/files#diff-535bb8ce6289b0a73b9e48df5f082035643112bf1bb868100dd2923d6bbbbdb5R704

tniessen:
@dmgorsky Looks great to me! I'll test it once https://github.com/XAMPPRocky/octocrab/pull/597 is released!

tniessen:
@dmgorsky Works like a charm as far as I can tell, thank you!

PR #600, opened 2024-03-07
Added `create_git_commit_object` endpoint

dmgorsky:
resolving #571

Had to modify `CommitAuthor` (models/repos.rs) to add `pub date: Option<DateTime<Utc>>`;
if it is unacceptable, then let's create separate struct


XAMPPRocky:
Thank you for your PR!

PR #603, opened 2024-03-10
Added test for `actions().create_workflow_dispatch()`

manchicken:
- 204 case
- 500 case


XAMPPRocky:
Thank you for your PR!

Issue #576, opened 2024-02-12
listed example for github app auth doesn't work but another provided method does


ericandoh:
Using the example in https://github.com/XAMPPRocky/octocrab/blob/main/examples/github_app_authentication_manual.rs
the below code does not work, returns `Failure from github: Error getting octocrab, GitHub: A JSON web token could not be decoded Documentation URL: https://docs.github.com/rest`

```
  // installation is gotten from octocrab as well earlier
  async fn get_octocrab(&self, installation: Installation) -> Result<Octocrab, ScanError> {
        let create_access_token = octocrab::params::apps::CreateInstallationAccessToken::default();
        println!("Processing installation: {}", installation.id.0);

        let access: octocrab::models::InstallationToken = self
            .octocrab
            .post(
                installation.access_tokens_url.as_ref().ok_or_else(|| {
                    ScanError::Github(
                        "Access token URL not found:".into(),
                        installation.id.to_string(),
                    )
                })?,
                Some(&create_access_token),
            )
            .await
            .map_err(|e| ScanError::Github("Errors at this call".into(), e.to_string()))?;
        //...Below code doesnt execute, error in former call
        octocrab::OctocrabBuilder::new()
            .personal_token(access.token)
            .build()
            .map_err(|e| {
                ScanError::Github(
                    "Failed to create octocrab instance with access token".into(),
                    e.to_string(),
                )
            })
```

However, this code works though, so I know my `self.octocrab` is set up with the correct app ID + secret. 
```
  async fn get_octocrab(&self, installation: Installation) -> Result<Octocrab, ScanError> {
        println!("Processing installation: {}", installation.id.0);

        let (myself, _secret) = self
            .octocrab
            .installation_and_token(installation.id)
            .await
            .map_err(|e| ScanError::Github("I'm an error 2".into(), e.to_string()))?;
        Ok(myself)
```

any idea why the below works but not the listed example?

ericandoh:
my octocrab:

`octocrab = { version = "^0.33.4", features = ["stream"] }`

ztj:
I believe this and a bunch of other problems were introduced by this pr: https://github.com/XAMPPRocky/octocrab/pull/562

[...]

PR #610, opened 2024-03-17
docs: Update manual github app authentication example

timonv:
When manually doing requests with octocrab, tokens are not forwarded to urls that contain an authority. This PR updates the example to reflect that requirement. 

Edit: makes sure the the documentation aligns with #576. Personally not against the current implementation, although the error could be better.


timonv:
My bad, assumed it was correct because because analyzer was not complaining :shrug:

XAMPPRocky:

PR #615, opened 2024-03-23
Change `Rate::reset` to `u64`

### LINKED ISSUE #614: Should `RateLimit::reset` be a `u64`?
Re this line: https://github.com/XAMPPRocky/octocrab/blob/131ddc02b2546f8c9863f98703fb92a3945ed8e8/src/models.rs#L959

The [GitHub REST API docs](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28) describe the corresponding header (`x-ratelimit-reset`) as "UTC epoch seconds".

This suggests that users should pass the value to [`Duration::from_secs`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_secs).

Since `Duration::from_secs` expects a `u64`, should this field be a `u64`?

smoelius:
Fixes #614

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -956,5 +956,5 @@ pub struct Rate {
     pub limit: usize,
     pub used: usize,
     pub remaining: usize,
-    pub reset: usize,
+    pub reset: u64,
 }


XAMPPRocky:
Thank you for your PR!

XAMPPRocky:
Though looking at this, probably all the fields should not be usize.

smoelius:
> Though looking at this, probably all the fields should not be usize.

I'm not sure, TBH. But since the the other fields are counts of "actions," `usize` seems reasonable.

XAMPPRocky:
Generally if it doesn't actually match to the size of a machine pointer, it shouldn't use the size types, it should be u32 or u64.

PR #630, opened 2024-05-01
feat: support permission on list_collaborators

### LINKED ISSUE #629: list_collaborators with permission
https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#list-repository-collaborators

It has a `permission` param. But this doesn't seem to exist at `ListCollaboratorsBuilder`.

tisonkun:
This closes #629.

### DIFF
--- a/src/api/repos/collaborators.rs
+++ b/src/api/repos/collaborators.rs
@@ -1,4 +1,16 @@
+use serde::{Deserialize, Serialize};
 use super::*;
+
+#[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+#[non_exhaustive]
+pub enum Permission {
+    Pull,
+    Triage,
+    Push,
+    Maintain,
+    Admin,
+}
 
 #[derive(serde::Serialize)]
 pub struct ListCollaboratorsBuilder<'octo, 'r> {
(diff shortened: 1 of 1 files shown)


src/api/repos/collaborators.rs
```diff
@@ -8,6 +8,8 @@ pub struct ListCollaboratorsBuilder<'octo, 'r> {
     per_page: Option<u8>,
     #[serde(skip_serializing_if = "Option::is_none")]
     page: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    permission: Option<String>,
```
tisonkun, line 12: Unsure if we should use an enum for the possible values instead.

I'm not quite familiar with enum serialization.

src/api/repos/collaborators.rs
```diff
@@ -31,6 +33,14 @@ impl<'octo, 'r> ListCollaboratorsBuilder<'octo, 'r> {
         self
     }
 
+    /// Filter collaborators by the permissions they have on the repository.
+    /// If not specified, all collaborators will be returned.
+    /// Can be one of: pull, triage, push, maintain, admin
+    pub fn permission(mut self, page: impl ToString) -> Self {
+        self.page = Some(page.to_string());
```
tisonkun, line 40:
```suggestion
    pub fn permission(mut self, permission: impl ToString) -> Self {
        self.permission = Some(permission.to_string());
```

src/api/repos/collaborators.rs
XAMPPRocky: Is there a known limited set of values it should be? Then it should use an `enum`. You can copy one of the other enums in the crate as an example.

tisonkun: Sounds reasonable. Let me update the code here in hours.

tisonkun: Updated.

tisonkun:
@XAMPPRocky I've run `cargo fmt --all` locally.

XAMPPRocky:
@tisonkun Sorry for the delay, could you rebase this PR?

XAMPPRocky:

Issue #632, opened 2024-05-01
Creating pr request with "head_repo" parameter


abrassel:
Is issuing a pr creation with the `head_repo` parameter set currently supported? If not, do you mind if I submit a pull request to add the feature?

PR #633, opened 2024-05-01
Add head repo to create pr

abrassel:
https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#create-a-pull-request supports adding `head_repo`, so support the same for create request.

Required for my use case.


abrassel:
https://github.com/XAMPPRocky/octocrab/issues/632

PR #636, opened 2024-05-06
Add head repo to create pr

abrassel:
https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#create-a-pull-request supports adding head_repo, so support the same for create request.

Required for my use case.
#632


XAMPPRocky:

PR #643, opened 2024-05-22
fix(builder): Change add_retry_config signature to match other Builder methods 

### LINKED ISSUE #642: Builder add_retry_config requires intermediate variable binding
Currently the add_retry_config function of OctocrabBuilder is defined with a different function signature than the others, requiring the builder to be bound to an intermediate variable before building to prevent move issues in the borrow checker.

I can't find any public usages of this function on GitHub, so I believe this odd behavior has gone unnoticed since the function was added.

cc @L1ghtman2k (we spoke about this offline)

Example usage:
image

davidsow:
Makes `add_retry_config()` adhere to standard set by other methods in the builder. 
This has been tested locally and appears to work exactly as expected given attached usage example.

Usage example: 
image

Closes #642


XAMPPRocky:

PR #650, opened 2024-06-10
Changing the user name from required to optional parameter

Kapple14:
Changing the user name parameter from required to optional following the official API guidelines. See https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28 for more


Kapple14:
Thanks for the fast approval.

XAMPPRocky:
I can't mer

XAMPPRocky:

PR #658, opened 2024-06-13
cargo fmt

maflcko:
Looks like this was forgotten before the latest merge commit.


XAMPPRocky:
Thank you for your PR! Would you mind also adding to the CI check so we can prevent it happening in the future?

XAMPPRocky:
Ah right, gotcha. Thank you for your PR!

PR #665, opened 2024-07-24
feat: allow sending non String payload with execute

### LINKED ISSUE #360: Sending non-String HTTP request bodies no longer supported?
#297 refactored various HTTP request APIs. 

New generic APIs for sending HTTP requests in lib.rs (e.g. `pub async fn execute(&self, request: http::Request<String>) -> Result<http::Response<hyper::Body>>` now mandate the use of `http::Request<String>`, effectively preventing the sending of binary bodies. In addition, building the request can transparently coerce binary bodies into String.

I am currently relying on Octocrab for uploading release assets and this change in behavior caused binary assets (represented as `Vec<u8>` to get encoded as their string/debug representation. See https://github.com/indygreg/python-build-standalone/issues/172.

Would it be possible to make `execute()` generic or for there to be a separate API to allow issuing of `http::Request<Vec<u8>>`? Essentially some way to issue HTTP requests with non-String bodies.

loispostula:
resolves #360

The approach I took is similar to the one outline here: https://github.com/XAMPPRocky/octocrab/issues/360#issuecomment-1547219725

### Drop of internal retry policy handler
Having to clone the request is a real challenge, as stated here: https://github.com/XAMPPRocky/octocrab/issues/360#issuecomment-1548524619 and https://github.com/hyperium/tonic/issues/733. I am in the opinion that such behaviour should not be inside of this library but in the client code, where they can use tools like [tryhard](https://docs.rs/tryhard/latest/tryhard/index.html) to achieve reliable retry mechanism.
- the `execute` API has changed, so anyone using it outside of the crate might have to impact their code

### DIFF
--- a/src/body.rs
+++ b/src/body.rs
@@ -1,0 +1,116 @@
+use http_body_util::BodyExt;
+
+use bytes::Bytes;
+use http_body::Frame;
+use snafu::{Backtrace, GenerateImplicitData};
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+type BoxBody = http_body_util::combinators::BoxBody<Bytes, crate::Error>;
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
+fn boxed<B>(body: B) -> BoxBody
+where
+    B: http_body::Body<Data = Bytes> + Send + Sync + 'static,
+    B::Error: Into<BoxError>,
+{
+    try_downcast(body).unwrap_or_else(|body| {
+        body.map_err(|e| crate::Error::Other {
+            source: e.into(),
+            backtrace: Backtrace::generate(),
+        })
+        .boxed()
+    })
+}
+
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR! 

> I am in the opinion that such behaviour should not be inside of this library but in the client code, where they can use tools like tryhard to achieve reliable retry mechanism.

I do not agree with this opinion, the octocrab client should be able of handling retries and throttling, just like the existing octokit client. I'm not going to merge a solution that removes this functionality. It may be a challenge but this is functionality that is worth preserving. If you're willing to make changes to this PR to support cloning streaming bodies, I'd be happy to merge it.

loispostula:
@XAMPPRocky I' ve updated the pr to allow cloning the request by changng `OctoBody<BoxBody>` to `OctoBody<Arc<RwLock<BoxBody>>>`

XAMPPRocky:

PR #679, opened 2024-08-15
Add `DeviceCodes::poll_until_available` method

mkrasnitski:
The polling behavior is specified by the GitHub API (such as waiting minimum 5 seconds between polls and what to do if you get a `slow_down` error). Therefore, it would be best for users not to implement this logic themselves as they may get it wrong. In principle it might even be a good idea to remove the `poll_once` method and just have `poll_until_available`.

I'm not sure if `await`-ing this function will actually block, but I'm assuming tokio will properly schedule other things while we're waiting to poll.


XAMPPRocky:
Thank you for your PR!

PR #682, opened 2024-08-24
Add missing fields in `CreateForkBuilder`

benpueschel:
As part of the [GitHub API](https://docs.github.com/en/rest/repos/forks?apiVersion=2022-11-28#create-a-fork--parameters), forking a repository supports two additional fields: `name`, and `default_branch_only`. 

These two parameters are currently missing in the `CreateForkBuilder` struct, and I think adding them is a good idea.


XAMPPRocky:

benpueschel:
Thank you :)

Issue #507, opened 2023-12-21
non-working example - issues paged list


panov-andy:
Hi,
I tried the example from the main page (README.md) and it returns only the first page of the issues.
Please, can you update the example?

```rust
let octocrab = octocrab::instance();
// Returns the first page of all issues.
let mut page = octocrab
    .issues("XAMPPRocky", "octocrab")
    .list()
    // Optional Parameters
    .creator("XAMPPRocky")
    .state(params::State::All)
    .per_page(50)
    .send()
    .await?;

// Go through every page of issues. Warning: There's no rate limiting so
// be careful.
loop {
    for issue in &page {
        println!("{}", issue.title);
    }
    page = match octocrab
        .get_page::<models::issues::Issue>(&page.next)
        .await?
    {
        Some(next_page) => next_page,
        None => break,
    }
}
```

panov-andy:
I tried to reduce the number of items_per_page and have got 3 pages only in total for public GitHub.

I also tried to run this example against corporate GitHub with debug logs enabled and have got `api/v3` twice for the next_page. 

`.base_uri("https://github.internal.corp/api/v3/")?`

```txt
[2023-12-21T18:02:48Z DEBUG octocrab] HTTP; http.method=GET http.url=https://github.internal.corp/api/v3/repos/kubernetes-live/auto-issues-live/issues?state=all&per_page=5 otel.name="HTTP" otel.kind="client"
[2023-12-21T18:02:48Z DEBUG octocrab] HTTP; http.method=GET http.url=https://github.internal.corp/api/v3/api/v3/repositories/215168/issues?state=all&per_page=5&page=2 otel.name="HTTP" otel.kind="client"
```

XAMPPRocky:
Thank you for your issue! I don't have the time currently to investigate, but would be happy to help review a fix for the example.

PR #684, opened 2024-08-30
Only add base_path if req_pandq does not contain it

wesemann:
* avoids adding it twice
* fixes issue with pagination not working due to base path added twice

Fixes https://github.com/XAMPPRocky/octocrab/issues/507


XAMPPRocky:

Issue #631, opened 2024-05-01
Support remove_assignees on issue API


tisonkun:
https://docs.github.com/en/rest/issues/assignees?apiVersion=2022-11-28#remove-assignees-from-an-issue

PR #686, opened 2024-08-30
Support `remove_assignees` on issue API

benpueschel:
Add support for removing assignees from an issue (https://docs.github.com/en/rest/issues/assignees?apiVersion=2022-11-28#remove-assignees-from-an-issue). 

Refs: #631


XAMPPRocky:
Thank you for your PR!

Issue #641, opened 2024-05-16
Return result instead of panic when returning an `installation` Octocrab instance


krishmas:
https://github.com/XAMPPRocky/octocrab/blob/7e280be901c378a1e162c37f909ea6590c6ff4ba/src/lib.rs#L981

Would be nicer to return an Error result (like other functions that return different types of Octocrabs do) instead of panicking so the application using doesn't immediately crash out

XAMPPRocky:
Thank you for your issue! A PR adding this would be welcome 🙂

PR #687, opened 2024-08-30
fix(installation)!: Return Result instead of panicking

benpueschel:
As #641 describes, returning a Result when calling `Octocrab::installation` is nicer than panicking outright. 

I'm definitely not happy with how I build and return the Errors - is there a better way to build Errors directly?

This PR changes the public API behavior, since `Octocrab::installation` is exposed to the user.

(closes: #641)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1008,20 +1008,27 @@ impl Octocrab {
     /// then obtain an installation ID, and then pass that here to
     /// obtain a new `Octocrab` with which you can make API calls
     /// with the permissions of that installation.
-    pub fn installation(&self, id: InstallationId) -> Octocrab {
+    pub fn installation(&self, id: InstallationId) -> Result<Octocrab> {
         let app_auth = if let AuthState::App(ref app_auth) = self.auth_state {
             app_auth.clone()
         } else {
-            panic!("Github App authorization is required to target an installation");
+            let source = std::io::Error::new(
+                std::io::ErrorKind::Other,
+                "Github App authorization is required to target an installation",
+            );
+            return Err(Error::Other {
+                backtrace: Backtrace::generate_with_source(&source),
+                source: Box::new(source),
+            });
         };
-        Octocrab {
+        Ok(Octocrab {
             client: self.client.clone(),
             auth_state: AuthState::Installation {
                 app: app_auth,
                 installation: id,
                 token: CachedToken::default(),
             },
-        }
+        })
     }
 
     /// Similar to `installation`, but also eagerly caches the installation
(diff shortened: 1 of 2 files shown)


benpueschel:
@XAMPPRocky Sorry for pinging you 😀 
Is there any better way to create custom errors in octocrab? 

I'm not really happy with the way I did it as a first quick draft:
```rust
let source = std::io::Error::new(
    std::io::ErrorKind::Other,
    "Github App authorization is required to target an installation",
);
return Err(Error::Other {
    backtrace: Backtrace::generate_with_source(&source),
    source: Box::new(source),
});
```

XAMPPRocky:
@benpueschel We use the snafu crate currently for custom errors, so you can use that.

https://docs.rs/snafu/latest/snafu/

benpueschel:
I may be missing something here - does snafu provide a way to conveniently build the `Error::Other` variant? 

If it doesn't, I could maybe add a new Error type (though I don't really like that - it's both not elegant and would introduce another breaking change for anyone matching on the octocrab Error type).

XAMPPRocky:
> If it doesn't, I could maybe add a new Error type (though I don't really like that - it's both not elegant and would introduce another breaking change for anyone matching on the octocrab Error type).

I would add an error variant. We should change the error enum to be `#[non_exhaustive]` to make that non breaking.

benpueschel:
Sounds good.

src/lib.rs
```diff
@@ -1008,20 +1008,25 @@ impl Octocrab {
...
-    pub fn installation(&self, id: InstallationId) -> Octocrab {
+    pub fn installation(&self, id: InstallationId) -> Result<Octocrab> {
         let app_auth = if let AuthState::App(ref app_auth) = self.auth_state {
             app_auth.clone()
         } else {
-            panic!("Github App authorization is required to target an installation");
+            return Err(Error::Installation {
+                backtrace: Backtrace::generate(),
+                source: error::InstallationError::new(
+                    "Github App authorization is required to target an installation",
```
XAMPPRocky, line 1018: I don't think this needs a message. I think you can just write as a display impl with snafu.

benpueschel: Yeah, you're right. I could also remove the entire `InstallationError` struct since we don't need a custom message. What do you say about this?
```rust
pub enum Error {
    ...
    #[snafu(display("Installation Error: Github App authorization is required to target an installation.\n\nFound at {}", backtrace))]
    Installation { backtrace: Backtrace },
    ...
}
```

I could shorten the message if it's too long, or omit the backtrace and make the entire variant a unit:

```rust
pub enum Error {
    ...
    #[snafu(display("GitHub App authorization is required to target an installation."))]
    Installation,
    ...
}
```

XAMPPRocky: I prefer the former. The backtrace is important

benpueschel: Alright. Do you prefer the first or second display message? I like the former because it keeps the style from all other error displays

XAMPPRocky: The former is good for now.

XAMPPRocky:
Thank you for your PR!

PR #695, opened 2024-09-17
Add `repository` to webhook events

InAnYan:
I've added a `repository: Repository` to `issues` and `issues_comments` events.

They are in the API, but not in `octocrab`


InAnYan:
When following the tutorial:

https://docs.rs/octocrab/latest/octocrab/index.html#github-webhook-application-support

This PR is wrong and unnecessary

PR #696, opened 2024-09-19
add additional webhook model fields

backwardspy:
i'm building a webhook handler and switched to octocrab's models today instead of my own hand-rolled models. there were a couple of fields i needed that weren't supported, so i've added them in and confirmed they seem to work as intended.

the naming of the repo change structs i added (owner/repo name) are a little confusing but the nesting is so deep i wasn't sure which way to go with it. let me know if you'd prefer something else!

if it helps at all, i have some example deliveries with these fields on them:

- repository.renamed: https://github.com/catppuccin/rockdove/blob/main/fixtures/repository_renamed.json#L3-L9
- repository.transferred: https://github.com/catppuccin/rockdove/blob/main/fixtures/repository_transferred.json#L3-L28


XAMPPRocky:
Thank you for your PR!

PR #697, opened 2024-09-21
fix(refs)!: remove `Reference::Commit` variant

benpueschel:
Closes: #608

Commits aren't valid refs and using the `Reference::Commit` variant will always return a 404, so this PR removes the variant altogether. 

This PR technically introduces a breaking change, but given that the feature doesn't work anyway, removing it should be safe.

### DIFF
--- a/src/params.rs
+++ b/src/params.rs
@@ -481,7 +481,6 @@ pub mod repos {
     pub enum Reference {
         Branch(String),
         Tag(String),
-        Commit(String),
     }
 
     impl Reference {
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR!

emiraga:
I believe you are wrong https://docs.github.com/en/repositories/working-with-files/using-files/downloading-source-code-archives#source-code-archive-urls

but workaround is simple
```
      octocrab
        .repos(&owner, &repo_name)
        .download_tarball(commit_hash)  // it's a String
```

PR #698, opened 2024-09-25
Implement some style improvements

benpueschel:
- Remove deprecated function calls in pr review tests, use the new functions
- Break-up large first paragraphs in doc comments
- Indent new lines on list elements in doc comments
- Call `assert!(result.is_err())` instead of `assert!(!result.is_ok())` in user blocks tests
- Implement `Display` trait instead of directly implementing `ToString` for `PullRequestTarget`


XAMPPRocky:
Thank you for your PR!

Issue #705, opened 2024-10-01
Make the Fields in models::timelines::Source Public


matilda-ray-kerria:
I am currently developing a program to manage GitHub issues and pull requests using the octocrab crate. While trying to extract information about related pull requests from the issue timeline, I noticed that the fields in `models::timelines::Source` are private.

```rust
use octocrab::{
    models::{timelines::Source, Event},
    Octocrab,
};

#[tokio::main]
async fn main() -> octocrab::Result<()> {
    let octocrab = Octocrab::builder()
        .personal_token("GITHUB_PERSONAL_TOKEN".to_string())
        .build()?;

    let page = octocrab
        .issues(OWNER, REPO)
        .list_timeline_events(1u64)
        .per_page(100)
        .send()
        .await?;

    for event in &page {
        if Event::CrossReferenced == event.event {
            if let Some(Source { issue, .. }) = &event.source {
                if issue.pull_request.is_some() {
                    println!(
                        "Found related pull request: #{} - {}",
                        issue.number, issue.title
                    );
                }
            }
        }
    }

    Ok(())
}
```

```
error[E0451]: field `issue` of struct `Source` is private
  --> src/main.rs:24:34
   |
24 |             if let Some(Source { issue, .. }) = &event.source {
   |                                  ^^^^^ private field

For more information about this error, try `rustc --explain E0451`.
error: could not compile `octocrab_example` (bin "octocrab_example") due to 1 previous error
```

https://github.com/XAMPPRocky/octocrab/blob/9fbf59c42a4c267204980955a1c70b80ee919964/src/models/timelines.rs#L97-L102

Since `Source` does not have getter methods defined for its fields, I cannot access these values. Would it be possible to change these fields to public? Alternatively, is there another method to achieve this?

XAMPPRocky:
Thank you for your issue! Seems like an oversight, if you'd like to make a PR making them public I can review it.

matilda-ray-kerria:
I'm sorry for forgetting to include the 'closes' keyword in my pull request.

PR #708, opened 2024-10-03
Make Fields in `models::timelines` Structs Public

matilda-ray-kerria:
#705
Changed the visibility of certain struct fields in `models::timelines` from private to public.

- Private fields in these structs make it difficult to access issue and pull request information from timeline events.
- Making these fields public allows developers to directly access the required data without encountering access errors.


XAMPPRocky:

Issue #706, opened 2024-10-02
Version `0.41` hard depends on `aws-lc-sys`, which cannot cross-compile to arm


theRookieCoder:
The new `octocrab 0.41` release results in a hard dependency on `aws-lc-sys` via `hyper-rustls` when using the `rustls` feature: https://github.com/XAMPPRocky/octocrab/blob/9fbf59c42a4c267204980955a1c70b80ee919964/Cargo.toml#L41

Unfortunately, limitations in cargo mean that it is impossible for upstream crates to not depend on `aws-lc-sys` without patching the dependency (`octocrab` in this case). It seems the solution is to turn off default features for that dependency, use neither `ring` nor `aws-lc`, and instead let the consumer decide.

PR #709, opened 2024-10-06
fix: rustls hard-dependency

benpueschel:
Remove hard-dependency on `aws-lc-sys` by adding two new features: `rustls-ring` and `rustls-aws-lc-rs`.

By default, `rustls-aws-lc-rs` is enabled (which is also a default feature for `hyper-rustls`), but the dependency on aws-lc-rs can be overriden by consumers of the library.

This should fix compilation on ARM, but we should maybe add a section about this in the README to let consumers know how to fix compilation issues for ARM.

Let me know if I should add a section to the README, or if there's better place to let user's know about this fix. 

Closes: #706

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ pretty_assertions = "1.4.0"
 graphql_client = "0.14.0"
 
 [features]
-default = ["follow-redirect", "retry", "rustls", "timeout", "tracing", "default-client", "rustls-aws-lc-rs"]
+default = ["follow-redirect", "retry", "rustls", "timeout", "tracing", "default-client", "rustls-ring"]
 
 follow-redirect = ["tower-http/follow-redirect"]
 retry = ["tower/retry", "futures-util"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,12 @@ http = "1.0.0"
 http-body = "1.0.0"
 http-body-util = "0.1.0"
 hyper = "1.1.0"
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! However shouldn't the default feature be the one that allows it to compile on ARM?

benpueschel:
> Thank you for your PR! However shouldn't the default feature be the one that allows it to compile on ARM?

Yeah, you're right. I'll make `ring` the default.

benpueschel:
Looks like ring somehow breaks Windows builds.

theRookieCoder:
It seems aws-lc does not support the Android triple (`aarch64-linux-android`) at all. But I do know ring does support it natively.

> Looks like ring somehow breaks Windows builds.

I'm not sure how that happened, but most depends (at least the ones I depend on) use ring by default when using rustls, I suppose because it was (de facto) standard for a while.

cpu:
> It seems aws-lc does not support the Android triple (aarch64-linux-android) at all.

That target triple is described as both supported and tested in the [aws-lc-rs platform docs](https://aws.github.io/aws-lc-rs/platform_support.html#platforms). You might consider opening an upstream issue if you're having trouble. In my experience they're very helpful for these sorts of problems.

benpueschel:
@XAMPPRocky do you think we could re-run the checks? I think some dependency briefly broke Windows builds, but that should be resolved now (hopefully 😄).

And I'd say as long as we don't know if aws-lc-rs supports ARM cross-compilation, going with ring by default should be fine.

Issue #649, opened 2024-06-09
Easily remove backtraces from error messages


theRookieCoder:
For code consuming `octocrab`, the error `Display` implementations are difficult to use because some of them spew out the entire backtrace. For example, if there is no internet connection, `octocrab::Error::Service` is returned with the message `Service Error: client error (Connect)`, I think this is enough for most cases. However, there is also a huge 50+ lines long backtrace that I can't easily remove unless I match every single error variant in a helper function.

I noticed all variants of the error enum have a backtrace field. I'm not sure how snafu works, but it would be nice if it was possible to isolate the backtrace. Even just a method for getting the message and backtrace separately would be sufficient.

benpueschel:
I suppose we could change the way we generate backtraces. the [Backtrace](https://doc.rust-lang.org/stable/std/backtrace/struct.Backtrace.html) struct has a different method `capture`, which only captures backtraces when the `RUST_BACKTRACE` or `RUST_LIB_BACKTRACE` environment variables are set. 

That would change the displayed message without much effort to something like 
```
HTTP Error: {message}

Found at disabled backtrace
```
instead of dumping the entire call stack, unless the user explicitly enables them.

If that's too ugly we may need to change the display attributes, buy I feel this is a good middle-ground.

@XAMPPRocky any thoughts on this?

PR #710, opened 2024-10-06
fix: don't capture backtraces by default

benpueschel:
Call `Backtrace::capture()` instead of the `Backtrace::generate()` method provided by snafu. This disables backtrace capturing unless either the `RUST_BACKTRACE` or `RUST_LIB_BACKTRACE` environment variables are set at runtime.

This completely disables backtrace capturing, not just displaying the backtrace though, so if the backtrace is too important to disable it completely by default, I could also refactor this PR to change the display behavior - maybe through a feature flag letting consumers decide if they want to work with backtraces or not.

Closes: #649


XAMPPRocky:
Thank you for your PR!

Issue #707, opened 2024-10-02
Version 0.41.0 doesn't compile


jhspetersson:
```
 Compiling octocrab v0.41.0
error: named argument never used                                                                                                                                                                                                
  --> C:\Users\User\.cargo\registry\src\index.crates.io-6f17d22bba15001f\octocrab-0.41.0\src\api\repos\release_assets.rs:98:20
   |
96 |             "/{}/releases/assets/{id}",
   |             -------------------------- formatting specifier missing
97 |             owner = self.parent.owner,
98 |             repo = self.parent.repo,
   |                    ^^^^^^^^^^^^^^^^ named argument never used

error[E0609]: no field `parent` on type `&ReleaseAssetsHandler<'_, '_>`                                                                                                                                                         
  --> C:\Users\User\.cargo\registry\src\index.crates.io-6f17d22bba15001f\octocrab-0.41.0\src\api\repos\release_assets.rs:97:26
   |
97 |             owner = self.parent.owner,
   |                          ^^^^^^ unknown field
   |
   = note: available field is: `handler`

error[E0609]: no field `parent` on type `&ReleaseAssetsHandler<'_, '_>`
  --> C:\Users\User\.cargo\registry\src\index.crates.io-6f17d22bba15001f\octocrab-0.41.0\src\api\repos\release_assets.rs:98:25
   |
98 |             repo = self.parent.repo,
   |                         ^^^^^^ unknown field
   |
   = note: available field is: `handler`

error[E0609]: no field `parent` on type `&ReleaseAssetsHandler<'_, '_>`                                                                                                                                                         
   --> C:\Users\User\.cargo\registry\src\index.crates.io-6f17d22bba15001f\octocrab-0.41.0\src\api\repos\release_assets.rs:110:28
    |
110 |         let request = self.parent.crab.build_request(builder, None::<&()>)?;
    |                            ^^^^^^ unknown field
    |
    = note: available field is: `handler`

error[E0609]: no field `parent` on type `&ReleaseAssetsHandler<'_, '_>`
   --> C:\Users\User\.cargo\registry\src\index.crates.io-6f17d22bba15001f\octocrab-0.41.0\src\api\repos\release_assets.rs:111:29
    |
111 |         let response = self.parent.crab.execute(request).await?;
    |                             ^^^^^^ unknown field
    |
    = note: available field is: `handler`

error[E0609]: no field `parent` on type `&ReleaseAssetsHandler<'_, '_>`
[...]

PR #690, opened 2024-09-07
Implement getting users and reopos by their respective IDs

InAnYan:
Closes #683.

Well, what I basically did is that I introduced `UserRef` and `RepoRef`.

They are enums with two variants: either full form (nickname, owner/name pair), or ids.

I implemented `std::fmt::Display` for them, so that it's easier to operate in API URLs.

No breaking changes!!

PR #711, opened 2024-10-06
fix: crate build issues with `stream` feature

benpueschel:
Fix a build issue introduced by #690 when building octocrab with the `stream` feature, which the pr checks didn't catch. 

Closes: #707

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,9 +16,9 @@ jobs:
     - run: sudo apt install llvm
       if: matrix.os == 'ubuntu'
     - run: rustup default ${{ matrix.channel }}
-    - run: cargo build --verbose --all-targets
+    - run: cargo build --all-targets
     - run: cargo test
-    - run: cargo build --verbose --all-targets -F stream
+    - run: cargo build --all-targets -F stream
     - run: cargo test -F stream
 
   wasm-build:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -13,6 +13,7 @@ jobs:
         os: [ubuntu, macos, windows]
     steps:
     - uses: actions/checkout@v2
+    - run: sudo apt install llvm
     - run: rustup default ${{ matrix.channel }}
     - run: cargo build --verbose --all-targets
     - run: cargo test
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,6 +14,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - run: sudo apt install llvm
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR! Would you mind adding a step in CI to check the stream feature?

benpueschel:
>   = note: /usr/bin/ld: final link failed: No space left on device
>           collect2: error: ld returned 1 exit status

Are the logs taking up too much space? Could that be the reason ld panics?

benpueschel:
This PR might genuinely be cursed...

benpueschel:
Okay, so there are two more solutions to this problem I can think of.

1. We could add an entire new job which is entirely responsible for checking the stream feature, meaning we (hopefully) don't pollute each runner with two different versions of every single target (one with the default, one with the stream features)
2. run `cargo clean` after successfully building the default feature-set. I really don't like this approach because it'd dramatically increase workflow run time.

Both of these approaches are compromises, and it'd be great if we could somehow universally extract more space - or tell cargo to delete any targets we don't need, without clearing the entire build cache, but that seems to be harder than anticipated. 

Does anyone have a better idea?

benpueschel:
Okay, so there are two more solutions to this problem I can think of.

1. We could add an entire new job which is entirely responsible for checking the stream feature, meaning we (hopefully) don't pollute each runner with two different versions of every single target (one with the default, one with the stream features)
2. run `cargo clean` after successfully building the default feature-set. I really don't like this approach because it'd dramatically increase workflow run time.

Both of these approaches are compromises, and it'd be great if we could somehow universally extract more space on every OS - or tell cargo to delete any targets we don't need, without clearing the entire build cache, but that seems to be harder than anticipated. 

Does anyone have a better idea?

Caellian:
Something like:
```yaml
jobs:
  build:
    runs-on: ${{ matrix.os }}-latest
    strategy:
      matrix:
        channel: [stable, beta, nightly]
        os: [ubuntu, macos, windows]
        features:
          - ""
          - "-F stream"
    steps:
    - uses: actions/checkout@v2
    - run: rustup default ${{ matrix.channel }}
    - run: cargo build --verbose --all-targets ${{ matrix.features }}
    - run: cargo test ${{ matrix.features }}
```

should work&trade;?

benpueschel:
That would definitely queue-up some jobs (GitHub only allows for 20 total concurrent jobs, a maximum of 5 of which can be MacOS runners: https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#usage-limits), but I think that's the most simple solution.

I'll try it out when I have the time to.

benpueschel:
Looks like we're finally passing all tests :)

PR #728, opened 2024-10-23
Custom executor for Wasm Support

omarabid:
This PR enables Octocrab to use a custom executor by using `Buffer::pair` instead of `Buffer::new` which relies on the Tokio runtime. It also change `std::time` to `webtime` crate which is a drop-in replacement for time that compiles on wasm.

This was tested on a wasm environment (specifically Cloudflare Workers)


XAMPPRocky:

omarabid:
Thanks for merging.

PR #740, opened 2024-12-05
Enable client reuse when authenticating with app user access tokens

kyle-leonhard:
This PR allows octocrab instances to be reused with different Github app user access tokens, similar the `installation` function enabling an instance to be reused with different installations of an app.

The motivation is to avoid creating a new connection every time a request needs made for a different user, user access token. I didn't see any good way to reuse or provide the connection created in `build` ([code](https://github.com/XAMPPRocky/octocrab/blob/fae5b089161f6e97a7cd1eb7b4c7c6aa2589ee61/src/lib.rs#L674))

Here's how it looks:

```rust
let octocrab = OctocrabBuilder::new().build()?;

// Later, use the client with many, varying access tokens.
let octocrab = octocrab.user_access_token(token)?
```

An alternative is to expose an `Octocrab::toBuilder` method to effectively clone the existing client with varying configuration. This looks hard given how configuration ends up embedded in service layers.

User access token documentation: https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-user-access-token-for-a-github-app

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -983,6 +983,11 @@ pub enum AuthState {
         /// The cached access token, if any
         token: CachedToken,
     },
+    /// Access token based authentication.
+    AccessToken {
+        /// The access token
+        token: SecretString,
+    },
 }
 
 pub type OctocrabService = Buffer<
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR!

Issue #525, opened 2024-01-18
[copilot] tag functionality


dmgorsky:
As of https://docs.github.com/en/rest/copilot?apiVersion=2022-11-28 ,
the following URL handlers in `octocrab` are missing:
- [x] /orgs/{org}/copilot/billing/selected_teams
- [ ] /orgs/{org}/members/{username}/copilot
- [x] /orgs/{org}/copilot/billing/seats
- [x] /orgs/{org}/copilot/billing
- [x] /orgs/{org}/copilot/billing/selected_users

PR #747, opened 2025-01-22
Partial implementation of Organization CoPilot APIs

Zarthus:
I am an enterprise administrator for our organization's GitHub instance, we use copilot at our organization. I've been developing software that uses these metrics and seats but the library did not yet support the endpoints I required.

This partially tackles https://github.com/XAMPPRocky/octocrab/issues/525

I have done a best effort of implementing the stuff I needed, I have left comments on things I did not [yet] implement. 

The testsuite tests against data GitHub gives but their samples are not entirely feature complete. It is possible I missed a nullable somewhere. Across this week, I will be testing against real data using my fork, if you'd like to wait until I've narrowed down the bugs I could find I'd advocate waiting until the testing checkboxes at the bottom is checked, but I'm also happy to submit follow up PRs to fixes if required.

### Implementation 

```rust
let org = client.orgs("foo");
let result: Page<Vec<CopilotUsage>> = org.copilot().usage().await?;
// ... 
```

- [x] copilot/billing
    - [ ] billing enum support 
- [x] copilot/billing/seats
- [x] copilot/metrics (PARTIAL)
    - [ ] complete api response (e.g. copilot chat, copilot merge request metrics) 
- [x] copilot/metrics/team (PARTIAL)
    - [ ] complete api response (e.g. copilot chat, copilot merge request metrics) 
- [x] copilot/usage (note: experimental API, but it has been around for a while)
- [x] copilot/usage/team (note: experimental API, but it has been around for a while)

https://docs.github.com/en/rest/copilot?apiVersion=2022-11-28

#### Implementation Notes

1: there is an interesting "API design" from GitHub where their json is effectively like this:

```
[
  { ... }
]
```

There is no reason I observed to wrap this into an one-element array, this means the code also looks kind of ugly: `usage` and `metrics` APIs return a `Vec<Vec<Struct>>`, because the documentation says they can be paginated but realistically nobody will. It's kind of awkward but I opted to translate literally and suggest library consumers make prettier wrappers around it. As the copilot API is changing from time to time and some are marked experimental (e.g. usage) trying to fix their issues.

2: 
Copilot usage APIs are experimental;

>This endpoint is in public preview and is subject to change.

#### Testing progress:

- [x] Tested against payloads from GitHub's documentation
- [ ] Tested with real data

Note that even if I tested against real data, it does not guarantee perfect infallibility. We are only human :)

### Disclosure

It is worth noting that some of this code was generated using GitHub Copilot, in particular the struct boilerplate of the models after I copy/pasted the API response as a comment. If the maintainers of this project would rather not have GenAI generated code in your project I encourage declining this PR. Those exact copy/paste's are visible in the test `.json` files.

### DIFF
--- a/src/api/orgs/copilot.rs
+++ b/src/api/orgs/copilot.rs
@@ -1,0 +1,77 @@
+use super::*;
+
+#[derive(serde::Serialize)]
+pub struct CopilotHandler<'octo, 'r> {
+    #[serde(skip)]
+    handler: &'r OrgHandler<'octo>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    per_page: Option<u8>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    page: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    since: Option<chrono::DateTime<chrono::Utc>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    until: Option<chrono::DateTime<chrono::Utc>>,
+}
+
+impl<'octo, 'r> CopilotHandler<'octo, 'r> {
+    pub fn new(handler: &'r OrgHandler<'octo>) -> Self {
+        Self {
+            handler,
+            per_page: None,
+            page: None,
+            since: None,
+            until: None,
+        }
+    }
+
+    /// Results per page (max 100).
+    pub fn per_page(mut self, per_page: impl Into<u8>) -> Self {
+        self.per_page = Some(per_page.into());
+        self
+    }
+
+    /// Page number of the results to fetch.
+    pub fn page(mut self, page: impl Into<u32>) -> Self {
+        self.page = Some(page.into());
+        self
+    }
+
+    // Show usage metrics since this date.
+    // This is a timestamp in ISO 8601 format (YYYY-MM-DDTHH:MM:SSZ).
+    // Maximum value is 28 days ago.
+    pub fn since(mut self, since: chrono::DateTime<chrono::Utc>) -> Self {
+        self.since = Some(since);
+        self
+    }
+
+    // Show usage metrics until this date.
+    // This is a timestamp in ISO 8601 format (YYYY-MM-DDTHH:MM:SSZ) and should not preceed the since date if it is passed.
+    pub fn until(mut self, until: chrono::DateTime<chrono::Utc>) -> Self {
+        self.until = Some(until);
+        self
+    }
+
+    // Retrieve copilot metrics for the entire organization
+    pub async fn metrics(
+        self,
+    ) -> crate::Result<crate::Page<Vec<crate::models::orgs_copilot::metrics::CopilotMetrics>>> {
+        let route = format!("/orgs/{org}/copilot/metrics", org = self.handler.owner);
+
+        self.handler.crab.get(route, Some(&self)).await
+    }
+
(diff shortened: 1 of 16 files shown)


src/api/orgs/copilot.rs
```diff
@@ -0,0 +1,117 @@
+use super::*;
+
+#[derive(serde::Serialize)]
+pub struct CopilotHandler<'octo, 'r> {
+    #[serde(skip)]
+    handler: &'r OrgHandler<'octo>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    per_page: Option<u8>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    page: Option<u32>,
```
Zarthus, line 10: I have not yet found a reason for pagination unless someone sets `per_page` to a lower amount. Feels very unnecessary but I kept it by the book: Usage and Metrics endpoints support pagination, billing does not. https://docs.github.com/en/rest/copilot/copilot-usage?apiVersion=2022-11-28

src/api/orgs/copilot.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub async fn metrics(
+        self,
+    ) -> crate::Result<crate::Page<Vec<crate::models::orgs_copilot::metrics::CopilotMetrics>>> {
+        let route = format!("/orgs/{org}/copilot/metrics", org = self.handler.owner);
+
+        self.handler.crab.get(route, Some(&self)).await
+    }
+
+    // Retrieve copilot metrics for a specific team within the organization
+    pub async fn metrics_team<T: ToString>(
```
Zarthus, line 65: The `_team` functions have a generic `<T: ToString>` so you can pass both a `&'a str` (in most cases, people only look for one organization and it's most likely `'static`) as well as a `String`. 

Most APIs in the codebase seemed to want a String but it didn't make sense to me to take something that is relatively constant, so this is more developer sugar than anything else.

Zarthus:
I tested this against real data and fixed the bugs/inconsistencies with their docs 👍🏻 some fields were nullable when docs suggested they weren't 😕

src/api/orgs/copilot.rs
```diff
@@ -0,0 +1,117 @@
...
+        let route = format!(
+            "/orgs/{org}/team/{team}/copilot/metrics",
+            org = self.handler.owner,
+            team = team.to_string()
+        );
+
+        self.handler.crab.get(route, Some(&self)).await
+    }
+
+    pub async fn usage(
```
Zarthus, line 78: RIP King.

https://github.blog/changelog/2024-10-30-github-copilot-metrics-api-ga-release-now-available/

XAMPPRocky:
Thank you for your PR! I will try to give it a review over the next week.

Zarthus:
@XAMPPRocky Thank you! I removed the `_usage` APIs as that API is now .. defunct. And they didn't do a great job informing the world it was deprecated. (they do now)

Zarthus:
@XAMPPRocky i get you're busy and that's fine (i just pull from my branch as a dep), I just added new endpoints to this MR and would appreciate re-approval to run the CI workflows just to make sure all is still green.

Zarthus:
@XAMPPRocky I think the changeset is relatively small, the majority is boilerplate and/or tests. It's also been running on my fork for weeks with good success. My main worries are making sure it fits the design pattern inline with the rest of octocrab.

Issue #749, opened 2025-02-01
Commits may fail to be deserialized because `sha` field may be null in `DiffEntry` if contents did not change


fiadliel:
When a commit includes file(s) that have no changes to the contents, the `sha` field is null. e.g. the file mode changed from 100644 to 10755.

When trying to parse the commit response, this error shows up:
```
Json { source: Error { path: Path { segments: [Map { key: "files" }, Seq { index: 0 }, Map { key: "sha" }] }, original: Error("invalid type: null, expected a string", line: 1, column: 4929) }
```

Looking at the code, this can only happen with the `sha` field of the `DiffEntry` struct. If I modify the code to make this field optional, the request works.

I note that this behavior actually differs from what GitHub claim here: https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit
It says that `sha` is a required field, but that doesn't appear to be the case in this one situation.

![Image](https://github.com/user-attachments/assets/1e6e6260-4d78-42d3-953b-2168023ad3ff)

PR #750, opened 2025-02-01
`sha` is optional in `DiffEntry` if file contents are unchanged

fiadliel:
When only metadata changes for a file, the `sha` field is null when requesting a commit. An example of this is if the file mode was changed.

This behavior is different from what the online documentation says, which is that this is a required field.

Fixes: #749


XAMPPRocky:

Issue #738, opened 2024-11-27
Rate limit errors when paginating using app authentication


djrodgerspryor:
I'm only part way through isolating this, but I'm making an issue in case anyone else is running into similar problems.

We recently upgraded to the most recent release (0.42.1) and we're hitting rate limits on one of our tools that lists a lot of github release pages.

There seem to be two problems:

1. Only the first request of a pagination is authenticated. See the two logged requests below:

```
Request {
    method: GET,
    uri: https://api.github.com/repos/BurntSushi/ripgrep/releases?,
    version: HTTP/1.1,
    headers: {
        "content-length": "0",
        "authorization": Sensitive,
        "user-agent": "octocrab",
    },
    body: OctoBody(
        RwLock {
            data: BoxBody,
            poisoned: false,
            ..
        },
    ),
}
Request {
    method: GET,
    uri: https://api.github.com/repositories/53631945/releases?page=2,
    version: HTTP/1.1,
    headers: {
        "content-length": "0",
        "user-agent": "octocrab",
    },
    body: OctoBody(
        RwLock {
            data: BoxBody,
            poisoned: false,
            ..
        },
    ),
}
```

This behaviour seems identical when using either `into_stream`, or explicitly calling `get_page` with the `next_page` property from the previous request. The above logs were generated via this code:

```rust
use futures::stream::TryStreamExt;

let releases: Vec<_> = octocrab
        .repos(organisation, repository)
        .releases()
        .list()
        .send()
        .await?
        .into_stream(&octocrab)
        .try_collect()
        .await?;
```

This means that our paginated requests are being subjected to the public rate limit.

2. There doesn't seem to be any automatic handling of the rate limit errors: I'm guessing that the relevant issue is https://github.com/XAMPPRocky/octocrab/issues/349 ? I tried enabling the new `retry`feature, but it doesn't handle 403 error or the rate limit header, so it's not very useful.

Handling 403s in the retry middleware would be easy to fix, but it's not the main problem, since the public rate limits are very tight for our use case.

djrodgerspryor:
I initially suspected https://github.com/XAMPPRocky/octocrab/pull/562 but it seems that the allowed_authorities check in `auth_header.rs` is passing, but `self.auth_header` is None when the request is sent, so it doesn't get sent.

djrodgerspryor:
For reference, this is our Octocrab initialization code:

```rust
octocrab::OctocrabBuilder::new()
[...]

PR #754, opened 2025-03-05
fix: insert auth headers for api.github.com

pfnsec:
`into_stream()` for pagination is currently broken. Consider the following case:

```rust
        let stream_client = client.clone();
        let stream = client
            .pulls(&owner, &repo)
            .list_files(issue_number)
            .await?
            .into_stream(&stream_client);

        pin!(stream);

        let mut objects = Vec::new();
        while let Some(diff) = stream.try_next().await? {
            tracing::info("{}", &diff.filename);
        }
```

When more than 30 files are modified in a pull request, [the Github API will return a response with a `link` header](https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2022-11-28#using-link-headers) containing a link to the next page of results.

Octocrab correctly resolves the `next` URL from this header when it exhausts the first page of results, and so try_next() will make another API call to Github of the form `https://api.github.com/repositories/{repo_id}/issues?page=2`.

However, because the authority of this URL is `api.github.com` and not `None`,[ Octocrab will not include auth headers with that request,](https://github.com/XAMPPRocky/octocrab/blame/ce8c885dc2701c891ce868c846fa25d32fd44ba2/src/lib.rs#L1710) and the request will fail.

This pull request alters this behaviour to include auth headers if the URL authority is `None` or `"api.github.com"`.

Fixes: 
https://github.com/XAMPPRocky/octocrab/issues/576
https://github.com/XAMPPRocky/octocrab/issues/738

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1703,15 +1703,26 @@ impl Octocrab {
         };
 
         if let Some(mut auth_header) = auth_header {
-            // Only set the auth_header if the authority (host) is empty (destined for
+            // Only set the auth_header if the authority (host) is api.github.com or empty (destined for
             // GitHub). Otherwise, leave it off as we could have been redirected
             // away from GitHub (via follow_location_to_data()), and we don't
             // want to give our credentials to third-party services.
-            if parts.uri.authority().is_none() {
-                auth_header.set_sensitive(true);
-                parts
-                    .headers
-                    .insert(http::header::AUTHORIZATION, auth_header);
+            match parts.uri.authority() {
+                None => {
+                    auth_header.set_sensitive(true);
+                    parts
+                        .headers
+                        .insert(http::header::AUTHORIZATION, auth_header);
+                }
+                Some(authority) if authority == "api.github.com" => {
+                    auth_header.set_sensitive(true);
+                    parts
+                        .headers
+                        .insert(http::header::AUTHORIZATION, auth_header);
+                }
+                Some(authority) => {
+                    tracing::warn!("Not inserting auth header: authority = {}", authority);
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


pfnsec:
@XAMPPRocky 
This is blocking our release to production - any chance I could get this looked at? 
Big thanks in advance.

XAMPPRocky:

pvandervelde:
@pfnsec Should this fix the issue for all API calls? I'm calling 

``` rust
let installations = octocrab
        .apps()
        .installations()
        .send()
        .await
        .unwrap_or(Page::<octocrab::models::Installation>::default())
        .take_items();

    let id = InstallationId(installation_id);
    let Some(installation_index) = installations.iter().position(|l| l.id == id) else {
        return Err(Error::FailedToFindAppInstallation(
            repository_owner.to_string(),
            source_repository.to_string(),
            installation_id,
        ));
    };

    let installation = &installations[installation_index];

    let create_access_token = CreateInstallationAccessToken::default();

    // Create an access token for the installation
    let access: InstallationToken = octocrab
        .post(
            installation.access_tokens_url.as_ref().unwrap(),
            Some(&create_access_token),
        )
        .await
        .map_err(|_| {
            Error::FailedToCreateAccessToken(
                repository_owner.to_string(),
                source_repository.to_string(),
                installation_id,
            )
        })?;

    // Use the API token
    let api_with_token = octocrab::OctocrabBuilder::new()
        .personal_token(access.token)
        .build()
        .unwrap();

  let app = api_with_token.current().app().await;
```

The last call returns an error `A JSON web token could not be decoded`, similar (but potentially not the same) as in #576. I have updated to 0.44. Note that the call works if you use an octocrab instance that has authenticated as an app but not as an installation 🤷

PR #756, opened 2025-03-10
fix(emails): `visibility` can be `null`

dndx:
According to the [`/user/emails` schema](https://docs.github.com/en/rest/users/emails?apiVersion=2022-11-28#list-email-addresses-for-the-authenticated-user), the `visibility` field in the response can be `null`. The `UserEmailInfo` made it mandatory, causing deserialization failures during requests against real GitHub APIs like this:

```
invalid type: null, expected string or map
```

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -1106,7 +1106,7 @@ pub struct UserEmailInfo {
     pub email: String,
     pub primary: bool,
     pub verified: bool,
-    pub visibility: EmailVisibilityState,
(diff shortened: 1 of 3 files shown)


omarabid:
@dndx This caught me by surprise and wasted me a good couple hours. Interestingly, this should always (?) fail because the no-reply email generated by github (I assume for all users?) has `null` for `visibility`.

XAMPPRocky:

PR #758, opened 2025-03-18
Add pagination query parameters for when listing pull requests files

beyera:
As the title indicates, this adds support for setting the paginated query parameters when listing pull request files.

This PR intentionally adheres to the same builder implementation convention within the existing module to maintain its scope and focus. However, I really think this project could benefit from something like [bon](https://bon-rs.com/) and I would potentially be interested in taking on that larger effort if desired. 

Note that this is a breaking change. I believe this is the right decision because the current implementation is quite limited, allowing a maximum of 30 files without any documentation indicating this restriction. As a result, anyone using it, including myself, may already unknowingly have a bug unless they happen to be aware of the 30-file cap. If necessary, I am willing to create a deprecated version of list_files API and the builder as a separate function.

### DIFF
--- a/src/api/pulls.rs
+++ b/src/api/pulls.rs
@@ -333,23 +333,24 @@ impl<'octo> PullRequestHandler<'octo> {
         self.crab.delete(route, Some(&map)).await
     }
 
-    /// List all `DiffEntry`s associated with the pull request.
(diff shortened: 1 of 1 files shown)


beyera:
Upon reflection, I realize that even without specifying the initial pagination query parameters, I can still retrieve all pages making this PR far less useful. My apologies for missing that!

PR #763, opened 2025-03-28
models: Add enum `EventInstallation::id()` helper

Enselic:
The only common field for both `EventInstallation::Full` and `EventInstallation::Minimal` is the `id`` field. Let's add a helper to make life easier for e.g. implementors of GitHub Apps that trigger on webhook events.


XAMPPRocky:
Thank you for your PR!

PR #764, opened 2025-03-28
examples/graphql_issues.rs: Update schema URL

Enselic:
The current link is 404:ed.

    $ curl --head -L https://docs.github.com/public/schema.docs.graphql
    HTTP/2 404
    content-length: 9

    $ curl --head -L https://docs.github.com/public/fpt/schema.docs.graphql
    HTTP/2 200
    content-length: 1262568


XAMPPRocky:
Thank you for your PR!

Issue #770, opened 2025-05-04
Name field missing in Author model


KaanMol:
Hey!

I saw that the name field is missing and [according to the docs](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user) it is in there (or null).
https://github.com/XAMPPRocky/octocrab/blob/53bb43545b650a6f68dc163e319f976e582303bd/src/models.rs#L406

If I add it in a PR, would you approve it?

XAMPPRocky:
Thank you for your issue! Yes, all corrections to models are more than welcome. 🙂

KaanMol:
> Can be closed now, I guess?

Yes! Forgot to close the issue, thanks!

PR #771, opened 2025-05-04
models: Add optional `name` field to `Author` struct

KaanMol:
The optional field `name` was missing as I mentioned in #770.
This has now been added, not all fields are added, think about `company`, `location`, and others as [listed here](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user).

I will leave it up to you if you find it valuable to have them added :)

I ran the tests, everything passed, as far as I saw, nothing should be broken from just this... Please let me know if I missed that lol

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -422,6 +422,7 @@ pub struct Author {
     pub received_events_url: Url,
     pub r#type: String,
     pub site_admin: bool,
+    pub name: Option<String>,
     pub patch_url: Option<String>,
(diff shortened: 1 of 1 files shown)


XAMPPRocky:

> This has now been added, not all fields are added, think about company, location, and others as [listed here](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user).

All fields should be available in octocrab. 🙂

PR #780, opened 2025-06-23
feat: Add support for exchanging oauth code for access token

### LINKED ISSUE #638: Support procuring a user access token from an oauth code
Support for calling `login/oauth/access_token` outside of just the DeviceCode flow would be useful for verifying users of GitHub apps that use oauth.

GitHub docs references:
 - Warning section of [Registering a GitHub App/About the setup URL](https://docs.github.com/en/apps/creating-github-apps/registering-a-github-app/about-the-setup-url)
 - [Generating a user access token for a GitHub App/Generating a user access token when a user installs your app](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-user-access-token-for-a-github-app#generating-a-user-access-token-when-a-user-installs-your-app)

mattiapitossi:
closes #638


src/auth.rs
```diff
@@ -258,6 +258,31 @@ impl DeviceCodes {
     }
 }
 
+/// Exchange a code for a user access token
+///
+/// see: https://docs.github.com/en/developers/apps/identifying-and-authorizing-users-for-github-apps
+pub async fn get_access_token(
```
XAMPPRocky, line 264: Is there a reason this is a free function as opposed to a impl method?

mattiapitossi: Hi @XAMPPRocky, TBH I didn't see any value in adding an additional structure as it was done for the device flow. In that case, it's useful as the params are received from another REST API. But if you see any case in which I'm not considering I can change this to be a method

mattiapitossi: Ok, it seems that in octokit is defined as a struct:

export type ExchangeWebFlowCodeGitHubAppOptions = {
  clientType: "github-app";
  clientId: string;
  clientSecret: string;
  code: string;
  redirectUrl?: string;
  request?: RequestInterface;
};

https://github.com/octokit/oauth-methods.js/blob/main/src/exchange-web-flow-code.ts

Also the redirectUri seems to be opt.

I'll update the MR and replicate this behavior

XAMPPRocky: Sorry I meant how does it look to get the access token? Like what does the actual method call look like if you used octokit?

mattiapitossi: There are two options (GitHub App is the recommended one):

```
export async function exchangeWebFlowCode(
  options: ExchangeWebFlowCodeOAuthAppOptions,
): Promise<ExchangeWebFlowCodeOAuthAppResponse>;

export async function exchangeWebFlowCode(
  options: ExchangeWebFlowCodeGitHubAppOptions,
): Promise<ExchangeWebFlowCodeGitHubAppResponse>;
```

https://github.com/octokit/oauth-methods.js/blob/main/src/exchange-web-flow-code.ts

src/auth.rs
```diff
@@ -301,3 +326,15 @@ struct PollForDevice<'a> {
...
+#[derive(Serialize)]
+struct ExchangeCodeForTokenParams<'a> {
+    /// Required. The client ID for your GitHub App.
+    client_id: &'a str,
+    /// Required. The client secret for your GitHub App.
+    client_secret: &'a str,
+    /// Required. The code received from the POST <https://github.com/login/oauth/authorize?client_id=CLIENT_ID> request.
+    code: &'a str,
+    // Strongly recommended. The URL in your application where users will be sent after authorization.
+    redirect_uri: &'a str,
```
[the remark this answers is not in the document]
mattiapitossi: Thanks for the review @kyle-leonhard. I also noticed that in octokit is defined as optional (https://github.com/XAMPPRocky/octocrab/pull/780#discussion_r2173718006)

kyle-leonhard (approved):
Looks great to me! I'll use it to replace my own, bespoke implementation! Thanks for doing it!

src/auth.rs
```diff
@@ -258,6 +258,58 @@ impl DeviceCodes {
     }
 }
 
+#[derive(serde::Serialize)]
+pub struct ExchangeWebFlowCodeBuilder<'octo, 'client_id, 'code, 'client_secret, 'redirect_uri> {
```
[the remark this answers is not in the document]
mattiapitossi: Thanks @kyle-leonhard, I was still relying on the octokit.js implementation which still doesn't use new fields. In the [docs](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-user-access-token-for-a-github-app#using-the-web-application-flow-to-generate-a-user-access-token), I also noticed that repository_id was also added so I added both

XAMPPRocky:
Thank you for your PR!

PR #781, opened 2025-07-02
fix: Commit search endpoint uses wrong type for deserialization

halfdanthemild:
In `models::repo::Commit`, `author` which is of type `CommitAuthor` (https://github.com/XAMPPRocky/octocrab/blob/main/src/models/repos.rs#L160) has a non-optional name field. But this was causing crashes for me with the error `Serde error: missing field 'name'` as Github does not return the author name here (`name` is also not present in the example response in the Github docs for this endpoint either: https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-commits). `models::commits::Commit` works for me locally.


XAMPPRocky:

PR #786, opened 2025-08-05
Add ability to update an existing label

illicitonion:
Also: urlencode label name when deleting

Spaces are legal in label names, but get errors if they're not escaped.
This feels like a responsibility of the thing constructing URLs, but I
can understand if it should actually be the caller's responsibility.


maflcko (approved):
lgtm, thx

XAMPPRocky:
Thank you for your PR!

PR #799, opened 2025-09-18
fix!: make CommitAuthor email nullable

augustoccesar:
### TL;DR issue
Currently JSON parsing breaks when handling Push Webhook events where the pusher is a bot without an email.

### Description
Currently, the Push webhook uses the `GitUserTime` (which internally use the flatten `CommitAuthor`) as the `Pusher`, which assumes that email is always present. After considering adding a new struct to represent the `Pusher`, I noticed that in the documentation it mentions that the email can be null for all the `Commiter`, `Pusher` and `Author`.

This was noticed while handling Push webhook events from [GitHub Merge Queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue), where the pusher is:
```json
"pusher": {
  "name": "github-merge-queue[bot]",
  "email": null
}
```

But also investigated and this seems to be the case for bots in general. As PRs where the pushes are done by [Renovate](https://docs.renovatebot.com/) also have the pusher as:
```json
"pusher": {
  "name": "renovate[bot]",
  "email": null
}
```

So, because of that, the suggestion here is to do the change directly on the `CommitAuthor`. This approach is a breaking change for the current API. Another solution if we want to avoid it being breaking would be to have a different type for the `Pusher` where the email is nullable/optional.

### Screenshots
Screenshot 2025-09-18 at 11 06 48
Screenshot 2025-09-18 at 09 50 51

### References
- https://docs.github.com/en/webhooks/webhook-events-and-payloads#push

### DIFF
--- a/src/api/repos.rs
+++ b/src/api/repos.rs
@@ -287,12 +287,12 @@ impl<'octo> RepoHandler<'octo> {
     ///     .branch("master")
     ///     .commiter(CommitAuthor {
     ///         name: "Octocat".to_string(),
-    ///         email: "octocat@github.com".to_string(),
+    ///         email: Some("octocat@github.com".to_string()),
     ///         date: None,
     ///     })
     ///     .author(CommitAuthor {
     ///         name: "Ferris".to_string(),
-    ///         email: "ferris@rust-lang.org".to_string(),
+    ///         email: Some("ferris@rust-lang.org".to_string()),
     ///         date: None,
     ///     })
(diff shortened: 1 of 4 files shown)


dmgorsky:
LGTM; [breaking] similar to #656 or #784

XAMPPRocky:
Thank you for your PR!

PR #804, opened 2025-09-27
Expose `digest` field in the `Asset` model

mominul:
GitHub releases now have automatically generated checksums in every asset. This PR exposes the `digest` field of the get release asset response.


XAMPPRocky (approved):

Sytten:
Any reason this is optional? The API seems to suggest this will always be present?

mominul:
@XAMPPRocky Thanks go to you for this awesome crate! ❤️

PR #812, opened 2025-10-10
fix: deserialize repo/generate response as Respository

jonaro00:
The official response schema is a Full Repository (https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#create-a-repository-using-a-template), and it is not clear why `application/vnd.github.baptiste-preview+json` is used here.

I have been using this patch in my application for a few days now with no issues.


XAMPPRocky:

Issue #808, opened 2025-10-06
Add version header


Sytten:
For future compatibility, the library should start sending the header `X-GitHub-Api-Version`.

PR #819, opened 2025-10-29
Feature/set http headers #808

dmgorsky:
proposal for #808

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -301,11 +301,11 @@ const GITHUB_BASE_URI: &str = "https://api.github.com";
 const GITHUB_BASE_UPLOAD_URI: &str = "https://uploads.github.com";
 
 /// This `include!` gives us pub const _SET_HEADERS_MAP: [(&str, &str)]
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -300,6 +300,15 @@ pub type Result<T, E = error::Error> = std::result::Result<T, E>;
 const GITHUB_BASE_URI: &str = "https://api.github.com";
 const GITHUB_BASE_UPLOAD_URI: &str = "https://uploads.github.com";
 
+/// This `include!` gives us pub const _SET_HEADERS_MAP: [(&str, &str)]
+/// generated from Cargo.toml `[package.metadata.github-api].set-headers` array, like
+/// ```
+/// [package.metadata.github-api]
+///
+/// set-headers = ["X-GitHub-Api-Version: 2022-11-28", ]
```
XAMPPRocky: Could we call this `headers` instead? TOML is declarative not imperative, so you wouldn't include a verb like "set"

PR #821, opened 2025-11-03
models: add missing issue event types

lschuermann:
This adds some issue event types encountered in the wild while trying to use [github-metadata-backup](https://github.com/0xB10C/github-metadata-backup/) on the [tock/tock](https://github.com/tock/tock) repository.


lschuermann:
I'm guessing that a more sustainable approach is to autogenerate these from GitHub's GraphQL API docs: https://docs.github.com/en/graphql/reference/objects

But ... this fixes an immediate issue now, so I'd love to see this merged eventually.

PR #824, opened 2025-11-10
`octocrab::Error`: don't include unconditional backtrace in Display impl

lschuermann:
Users of this crate cannot currently get an error description without the backtrace included. This may be undesirable for a number of reasons, and is inconsistent with the examples in the `snafu` crate (whose Display implementations generally do not include a backtrace). Users who want a backtrace to be printed can still do so for an octocrab Error `e` by writing:

    snafu::ErrorCompat::backtrace(&e)

This was requested as part of a pull request to a project using this crate: https://github.com/0xB10C/github-metadata-backup/pull/11#discussion_r2509355952

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -43,7 +43,7 @@ pub enum Error {
         backtrace: Backtrace,
     },
 
-    #[snafu(display("HTTP Error: {}\n\nFound at {}", source, backtrace))]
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR!

PR #829, opened 2025-12-01
fix: incorrect path for followers and followees

ssiyad:
```rust
async fn followers() -> Vec<Follower> {
    let octo = octocrab::instance();
    octo.users(USER)
        .followers()
        .send()
        .await
        .unwrap()
        .into_stream(&octo)
        .try_collect::<Vec<Follower>>()
        .await
        .unwrap()
}
```
This results in a 404 error because the path being called is `/users/users/<username>/followers`. This PR removes the extra `/users` part. Same for following.


ssiyad:
@dmgorsky done

PR #834, opened 2025-12-15
fix!: add a cfg to make the crypto backend for jwt configurable

CommanderStorm:
Since 0.48 / https://github.com/XAMPPRocky/octocrab/pull/814 the crypto backend of octocrab is hardcoded to be a mix of rust_crypto (for jwt) and the choice for rustls.

This is not ideal, since this probibits choosing the `jsonwebtoken/aws_lc_rs` feature in another crate.
I would like to do this as this way I only need to compile multiple crypto librarys.

The alternative of letting `rustls-aws-lc-rs`/.. do this choice does not work since `rust-crypto` is not yet recomended for production use, which means that adding it as a cfg is likely not a good idea.

> [!IMPORTANT]
> this is a breaking change due to the case that someone might have `default-features = false`

### DIFF
--- a/src/models/events/payload/push.rs
+++ b/src/models/events/payload/push.rs
@@ -1,5 +1,6 @@
 use serde::{Deserialize, Serialize};
 
+use super::Commit;
 use crate::models::PushId;
 
 /// The payload in a [`super::EventPayload::PushEvent`] type.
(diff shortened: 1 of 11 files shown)


dmgorsky:
```
/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/jsonwebtoken-10.2.0/src/lib.rs:14:1
compile_error!("at least one of the features \"rust_crypto\" or \"aws_lc_rs\" must be enabled");
```
@CommanderStorm

PR #848, opened 2025-12-30
fix: resolve docs.rs build failure

michalskalski:
The docs.rs build was failing [1] because `all-features = true` enables both `jwt-rust-crypto` and `jwt-aws-lc-rs`, which are mutually exclusive (jsonwebtoken crate forbids enabling both simultaneously [2]).

Fixes regression introduced in #834.

[1] https://docs.rs/crate/octocrab/latest/builds/2760766
[2] https://github.com/Keats/jsonwebtoken/blob/53a3fc25b48be0af5fdab63b5698b2a24f59b3ef/src/lib.rs#L8-L11


XAMPPRocky:
Thank you for your PR!

PR #870, opened 2026-04-01
Expose OctoBody

nataliejameson:
In #869, OctoBody was used in a public trait, but is not itself visible. It's marked pub, so I'm assuming that re-exporting it is okay.


XAMPPRocky (approved):

XAMPPRocky:
Thank you for your PR!

PR #879, opened 2026-04-28
Use PUT not PATCH for pull request reviews

nataliejameson:
I was getting 404s when trying to update the body of a pull request review. Looks like https://docs.github.com/en/rest/pulls/reviews?apiVersion=2026-03-10#update-a-review-for-a-pull-request says it's a PUT not PATCH.


XAMPPRocky:
Thank you for your PR!

PR #883, opened 2026-05-04
remove the either dependency

marcoieni:
Substitute the Either type with an enum, which clearer names wrt left and right.

The `poll_once` function is public, so this is a breaking change.

## AI disclosure

I used GPT5.5 low thinking with the codex harness to generate this code change. I changed nothing from the generated code because it looks good to me.

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -2,7 +2,6 @@
 
 use crate::Result;
 use crate::{models::AppId, Octocrab};
-use either::Either;
 use jsonwebtoken::{Algorithm, EncodingKey, Header};
 use secrecy::{ExposeSecret, SecretString};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


src/auth.rs
```diff
@@ -223,8 +230,8 @@ impl DeviceCodes {
             )
             .await?;
         Ok(match poll {
-            TokenResponse::Ok(k) => Either::Left(k),
-            TokenResponse::Continue { error } => Either::Right(error),
+            TokenResponse::Ok(auth) => PollResult::Authenticated(auth),
+            TokenResponse::Continue { error } => PollResult::Continue(error),
```
marcoieni: I guess an alternative is returning TokenResponse directly.

XAMPPRocky: Yeah, I think exposing TokenResponse makes more sense.

XAMPPRocky:
Thank you for your PR!

Issue #224, opened 2022-07-09
Octocrab doesn't compile with Yew (for me)


NicMcPhee:
I was hoping to use Octocrab on a Yew-based project that I'm working on, but when I added the octocrab dependency to my `Cargo.toml` file, octocrab failed to compile.

I created a small demo app that illustrates the issue: https://github.com/NicMcPhee/octocrab_test This is about the simplest possible Yew app, and if you uncomment the octocrab dependency in `Cargo.toml` and run `trunk serve`, you get a whole pile of compilation errors when it's trying to build the octocrab code (see below).

I'm guessing it's some kind of dependency conflict, but I really don't know.

Yew + Octocrab seems like an "obvious" combination, so it was definitely a bummer that this didn't work. Any suggestions or advice would definitely be appreciated.

### Rust versions, etc.

`rustup show` provides the following output if that's helpful:

```text
Default host: x86_64-apple-darwin
rustup home:  /Users/mcphee/.rustup

installed targets for active toolchain
--------------------------------------

wasm32-unknown-unknown
x86_64-apple-darwin

active toolchain
----------------

stable-x86_64-apple-darwin (default)
rustc 1.62.0 (a8314ef7d 2022-06-27)
```

### The error text

```text
   Compiling octocrab v0.16.0
error[E0599]: no method named `user_agent` found for struct `ClientBuilder` in the current scope
   --> /Users/mcphee/.cargo/registry/src/github.com-1ecc6299db9ec823/octocrab-0.16.0/src/lib.rs:338:14
    |
338 |             .user_agent("octocrab")
    |              ^^^^^^^^^^ method not found in `ClientBuilder`

error[E0599]: no method named `user_agent` found for struct `ClientBuilder` in the current scope
   --> /Users/mcphee/.cargo/registry/src/github.com-1ecc6299db9ec823/octocrab-0.16.0/src/lib.rs:432:18
    |
432 |                 .user_agent("octocrab")
    |                  ^^^^^^^^^^ method not found in `ClientBuilder`

error: future cannot be sent between threads safely
  --> /Users/mcphee/.cargo/registry/src/github.com-1ecc6299db9ec823/octocrab-0.16.0/src/from_response.rs:11:80
   |
11 |       async fn from_response(response: reqwest::Response) -> crate::Result<Self> {
   |  ________________________________________________________________________________^
12 | |         let text = response.text().await.context(crate::error::HttpSnafu)?;
13 | |
[...]

PR #847, opened 2025-12-29
Add WebAssembly (WASM) support

sanks011:
This commit adds comprehensive WASM support to octocrab, enabling its use in browser-based applications with frameworks like Yew, Leptos, and Dioxus.

Changes:
- Add new wasm module with ReqwestTowerService for WASM compatibility
- Implement wasm_builder() function for easy WASM initialization
- Add conditional WASM dependencies (wasm-bindgen-futures, reqwest, thiserror)
- Create wasm_example.rs demonstrating WASM usage
- Update README.md with WASM usage documentation
- Update CHANGELOG.md with new feature

Fixes #224

PR #885, opened 2026-05-06
Add wasm client service

VoltVoks:
## Summary
- add a browser-compatible reqwest-backed service for `wasm32-unknown-unknown`
- make the wasm service available automatically on wasm targets while keeping the native service stack unchanged
- document the wasm setup, add an example, and keep CI building the wasm path

This is a smaller current-main alternative to #847, which is currently conflicting.

Fixes #224.

## Testing
- `cargo fmt --all -- --check`
- `cargo check`
- `cargo build --target=wasm32-unknown-unknown --no-default-features -F jwt-rust-crypto`
- `cargo test`
- `cargo clippy --example wasm_client` (passes with existing clippy warnings)
- external consumer check: `cargo check --target wasm32-unknown-unknown` against this branch with `default-features = false` and `features = ["jwt-rust-crypto"]`

### DIFF
--- a/examples/wasm_client.rs
+++ b/examples/wasm_client.rs
@@ -1,0 +1,36 @@
+#[cfg(target_arch = "wasm32")]
+use std::sync::Arc;
+
+#[cfg(target_arch = "wasm32")]
+use http::{header::USER_AGENT, Uri};
+#[cfg(target_arch = "wasm32")]
+use octocrab::service::middleware::base_uri::BaseUriLayer;
+#[cfg(target_arch = "wasm32")]
+use octocrab::service::middleware::extra_headers::ExtraHeadersLayer;
+#[cfg(target_arch = "wasm32")]
+use octocrab::service::wasm::ReqwestService;
+#[cfg(target_arch = "wasm32")]
+use octocrab::{AuthState, Octocrab, OctocrabBuilder};
+
+#[cfg(target_arch = "wasm32")]
+fn build_octocrab() -> octocrab::Result<Octocrab, std::convert::Infallible> {
+    OctocrabBuilder::new_empty()
+        .with_service(ReqwestService::new())
+        .with_layer(&BaseUriLayer::new(Uri::from_static(
+            "https://api.github.com",
+        )))
+        .with_layer(&ExtraHeadersLayer::new(Arc::new(vec![(
+            USER_AGENT,
+            "octocrab".parse().unwrap(),
+        )])))
+        .with_auth(AuthState::None)
+        .build()
+}
+
+#[cfg(target_arch = "wasm32")]
+fn main() {
+    let _octocrab = build_octocrab().expect("build wasm client");
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+fn main() {}
--- a/examples/wasm_client.rs
+++ b/examples/wasm_client.rs
@@ -1,36 +1,31 @@
-#[cfg(target_arch = "wasm32")]
-use std::sync::Arc;
+cfg_if::cfg_if! {
+    if #[cfg(target_arch = "wasm32")] {
+        use std::sync::Arc;
 
-#[cfg(target_arch = "wasm32")]
-use http::{header::USER_AGENT, Uri};
-#[cfg(target_arch = "wasm32")]
-use octocrab::service::middleware::base_uri::BaseUriLayer;
-#[cfg(target_arch = "wasm32")]
-use octocrab::service::middleware::extra_headers::ExtraHeadersLayer;
-#[cfg(target_arch = "wasm32")]
-use octocrab::service::wasm::ReqwestService;
-#[cfg(target_arch = "wasm32")]
-use octocrab::{AuthState, Octocrab, OctocrabBuilder};
+        use http::{header::USER_AGENT, Uri};
+        use octocrab::service::middleware::base_uri::BaseUriLayer;
+        use octocrab::service::middleware::extra_headers::ExtraHeadersLayer;
+        use octocrab::service::wasm::ReqwestService;
+        use octocrab::{AuthState, Octocrab, OctocrabBuilder};
 
-#[cfg(target_arch = "wasm32")]
-fn build_octocrab() -> octocrab::Result<Octocrab, std::convert::Infallible> {
-    OctocrabBuilder::new_empty()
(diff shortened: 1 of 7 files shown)


XAMPPRocky:
Thank you for your PR! Why have this be an explicit feature, why not just have it be the default for wasm32? If the project can't compile without the `wasm-client` feature it's probably better to just cfg define the dependencies.

VoltVoks:
Thanks, agreed. I removed the explicit `wasm-client` feature and made the reqwest-backed wasm service available automatically under `cfg(target_arch = "wasm32")`.

The existing `wasm-build` job now covers that path, and all CI checks are green on the updated commit. I also re-ran the local wasm consumer check without the `wasm-client` feature.

examples/wasm_client.rs
```diff
@@ -0,0 +1,36 @@
...
+        )))
+        .with_layer(&ExtraHeadersLayer::new(Arc::new(vec![(
+            USER_AGENT,
+            "octocrab".parse().unwrap(),
+        )])))
+        .with_auth(AuthState::None)
+        .build()
+}
+
+#[cfg(target_arch = "wasm32")]
```
XAMPPRocky: This is too many `cfg`s, I think it would be better to use `cfg_select!` here

src/lib.rs
```diff
@@ -307,7 +332,9 @@ compile_error!("at least one of the features \"jwt-rust-crypto\" and feature \"j
 /// A convenience type with a default error type of [`Error`].
 pub type Result<T, E = error::Error> = std::result::Result<T, E>;
 
+#[cfg(feature = "default-client")]
 const GITHUB_BASE_URI: &str = "https://api.github.com";
+#[cfg(feature = "default-client")]
 const GITHUB_BASE_UPLOAD_URI: &str = "https://uploads.github.com";
```
XAMPPRocky, line 338: Why are these consts guarded? They should be used with both clients no?

src/lib.rs
```diff
@@ -1128,6 +1265,17 @@ impl Octocrab {
         }
     }
 
+    /// Creates a new `Octocrab` with a custom executor.
+    #[cfg(target_arch = "wasm32")]
+    fn new_with_executor<S>(service: S, auth_state: AuthState, _executor: Executor) -> Self
```
XAMPPRocky: This is too many `cfg`s here, I would colocate everything into a single impl block with a `cfg` rather than doing it around individual methods.

XAMPPRocky (changes requested):

VoltVoks:
I pushed an update after the review that cleans up the cfg layout in the wasm example, leaves the GitHub base URI consts shared, and moves the wasm-specific builder/Octocrab methods into cfg-scoped impl blocks.

All current checks are green, including wasm-build and clippy. Could you take another look when you have a chance?

src/lib.rs
```diff
@@ -226,31 +226,50 @@ use api::repos::RepoRef;
...
 use http_body_util::combinators::BoxBody;
 use http_body_util::BodyExt;
+#[cfg(feature = "default-client")]
 use service::middleware::auth_header::AuthHeaderLayer;
-use service::middleware::cache::{CacheStorage, HttpCacheLayer};
+use service::middleware::cache::CacheStorage;
+#[cfg(feature = "default-client")]
+use service::middleware::cache::HttpCacheLayer;
+#[cfg(target_arch = "wasm32")]
+use std::cell::RefCell;
```
XAMPPRocky, line 240: Instead of cfging the imports. I would just use their full path where they are used. They are used throughout the codebase, so using the full path reduces the amount of divergences.

src/lib.rs
XAMPPRocky: I'm still confused, why would these be dead in the wasm32 version?

examples/wasm_client.rs
```diff
@@ -0,0 +1,31 @@
+cfg_if::cfg_if! {
```
XAMPPRocky, line 1: Why `cfg_if` and not `cfg_select` as suggested?

dairon2:
I took a pass at a small follow-up for the current review feedback and opened it against this branch here: https://github.com/VoltVoks/octocrab-bounty-work/pull/1

It addresses two of the maintainer comments directly:
- switches the wasm example from `cfg_if!` to `cfg_select!`
- removes several target-specific imports in `src/lib.rs` by using fully-qualified paths at the call sites, reducing cfg divergence around the service wrapper types

Kept intentionally small so it can be reviewed/applied quickly. I can keep helping with the remaining cfg consolidation if that is useful.

keshav55:
I opened a small follow-up against the active branch here: https://github.com/VoltVoks/octocrab-bounty-work/pull/4

It focuses only on the remaining default URI/dead-code review point: adds a wasm-only default builder path that reuses the same GitHub API/upload URI constants and default header/base-URI stack, while avoiding overlap with dairon2#1. Validation passed locally: fmt, wasm check, native check, and clippy.

dairon2:
Thanks, this looks complementary to VoltVoks#1. My follow-up intentionally stayed focused on the example `cfg_select!` cleanup and reducing cfg-scoped imports via fully-qualified paths. This PR seems to cover the remaining default URI / builder-path point, so both should be able to land cleanly. I’m happy to rebase or adjust #1 depending on which follow-up gets merged first.

dhiyaancnirmal:
I opened a focused follow-up against the active bounty branch here: https://github.com/VoltVoks/octocrab-bounty-work/pull/5

It fixes the remaining wasm example dependency-surface issue I hit with:

```bash
cargo check --target wasm32-unknown-unknown --no-default-features --features jwt-rust-crypto --example wasm_client
```

The patch keeps native behavior intact, moves native-only dev dependencies out of the wasm target, disables default `hyper`/`hyper-util` features for the wasm check path, and reuses the shared GitHub base URI constants in the example.

Validation passed locally:

```bash
cargo fmt --check
cargo check
cargo check --target wasm32-unknown-unknown --no-default-features --features jwt-rust-crypto --example wasm_client
```

Disclosure: this follow-up was prepared with AI assistance and locally reviewed/tested before submission.

XAMPPRocky:
It seems like all of these contributions are primarily AI driven, and not driven by people, so I'm going to close this PR along with all of the spam sub-PRs. Please go contribute somewhere else.

PR #891, opened 2026-05-14
fix: Octocrab doesn't compile with Yew (for me)

asiokun:
Fixes #224

Compiles cleanly. The changes are:

**`src/from_response.rs`** (lines 7-8, 15-16): Replaced both `#[async_trait::async_trait]` attributes on the `FromResponse` trait definition and its blanket impl with conditional `cfg_attr` pairs — using the standard `async_trait` on non-wasm32 and `async_trait(?Send)` on wasm32.

**`src/page.rs`** (line 180-181): Same change for `Page<T>`'s `FromResponse` impl.

**`src/models/repos.rs`** (line 262-263): Same change for `ContentItems`'s `FromResponse` impl.

**Root cause**: `async_trait` by default generates futures with a `+ Send` bound. On `wasm32-unknown-unknown` (used by Yew/Trunk), many types backed by JavaScript objects (e.g. `reqwest::Response` in the older version, or any JS-bridged type) are not `Send`. The `?Send` variant relaxes this requirement so futures don't need to be `Send`, which is correct for the single-threaded WASM runtime.

### DIFF
--- a/src/from_response.rs
+++ b/src/from_response.rs
@@ -4,14 +4,16 @@ use http_body_util::BodyExt;
 use snafu::ResultExt;
 
 /// A trait for mapping from a `http::Response` to an another type.
-#[async_trait::async_trait]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
As mentioned in #224 I am not existing contributions from new contributors for this feature.

PR #892, opened 2026-05-16
fix: resolve WASM Send trait violation in FromResponse (#224)

NguyenTrung-3107:
This PR fixes the compilation error on WASM where futures returned by sync_trait are not Send because the underlying body type B (often from 
eqwest on WASM) does not implement Send.

The fix uses cfg_if to:
1. Use #[async_trait(?Send)] on WASM targets.
2. Remove the B: Send bound in the where clause on WASM targets.

This allows octocrab to be used in WASM environments like Yew or other browser-based async runtimes without violating trait bounds.

### DIFF
--- a/src/from_response.rs
+++ b/src/from_response.rs
@@ -3,23 +3,52 @@ use http_body::Body;
 use http_body_util::BodyExt;
 use snafu::ResultExt;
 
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
As mentioned in #224 I'm not accepting contributions from new contributors for this feature.

Issue #897, opened 2026-05-18
`>=0.50.0` fails when parsing payloads with `pull-request-minimal` and have type as `PullRequest`


augustoccesar:
### Description

Since we are using the same PullRequest on multiple places, the changes on https://github.com/XAMPPRocky/octocrab/pull/873 broke parsing [`CheckRun`](https://github.com/XAMPPRocky/octocrab/blob/fa65c61f676eb1971e1819b6304d42706aaf14a0/src/models/checks.rs#L18) since it has a vec of `PullRequest`, but it is actually a `#/components/schemas/pull-request-minimal` where it does not have `node_id`, which was marked as required by that PR.

The newest added `SimplePullRequest` (https://github.com/XAMPPRocky/octocrab/pull/886) also cannot be used there since it also has a required `node_id`.

I can take a look into creating a PR for this, but wanted to open for discussion first before simply adding even one more representation of a `PullRequest` (`PullRequestMinimal`?). Wondering if is worth taking a look at the overall definitions and align them with the OpenAPI spec. This is also currently a issue for different representations of `Commit`.

### Payloads References

(OpenAPI reference https://raw.githubusercontent.com/github/rest-api-description/refs/heads/main/descriptions/api.github.com/api.github.com.2022-11-28.json)

Payload for the check run completed:
```json
"webhook-check-run-completed": {
  "title": "Check Run Completed Event",
  "type": "object",
  "properties": {
    "action": {
      "type": "string",
      "enum": [
        "completed"
      ]
    },
    "check_run": {
      "$ref": "#/components/schemas/check-run-with-simple-check-suite"
    },
    "installation": {
      "$ref": "#/components/schemas/simple-installation"
    },
    "enterprise": {
      "$ref": "#/components/schemas/enterprise-webhooks"
    },
    "organization": {
      "$ref": "#/components/schemas/organization-simple-webhooks"
    },
    "repository": {
      "$ref": "#/components/schemas/repository-webhooks"
    },
    "sender": {
      "$ref": "#/components/schemas/simple-user"
    }
  },
  "required": [
    "check_run",
    "repository",
    "sender"
  ]
},
```

`#/components/schemas/check-run-with-simple-check-suite`
```json
"check-run-with-simple-check-suite": {
  "title": "CheckRun",
  "description": "A check performed on the code of a given code change",
  "type": "object",
  "properties": {
    "pull_requests": {
      "type": "array",
      "items": {
[...]

PR #902, opened 2026-05-23
made PullRequest to be able to accept SimplePullRequest payload from GitHub

dmgorsky:
generalized for #897


manchicken:
Do we do anything to call out breaking changes? These changes look great, but they do look breaking.

XAMPPRocky:
Thank you for your PR! 

> Do we do anything to call out breaking changes? These changes look great, but they do look breaking.

If you suffix your semantic commit with a `!`, e.g. `feat!:` It will automatically be marked as breaking in the changelog.

PR #903, opened 2026-05-25
Added GitHub async SBOM API endpoints.

manchicken:
- Generate an SBOM report
- Poll for the report after it's been generated
- Fetch and deserialize the report once it's ready
- Added tests
- Added an example
- Added docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
-
-- Added `sbom` API for fetching GitHub SBOM information using asynchronous endpoints.
-
 ## [0.51.0](https://github.com/XAMPPRocky/octocrab/compare/v0.50.0...v0.51.0) - 2026-05-09
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `sbom` API for fetching GitHub SBOM information using asynchronous endpoints.
+
 ## [0.51.0](https://github.com/XAMPPRocky/octocrab/compare/v0.50.0...v0.51.0) - 2026-05-09
 
 ### Fixed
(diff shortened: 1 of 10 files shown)


manchicken:
Please note: all of my commits are GnuPG-signed.

manchicken:
Ok, @XAMPPRocky, I think this one is ready to go. Let me know if you spot any issues.

I certify that I created and submitted this PR without any code or documentation contributions from generative AI of any kind.

src/api/repos/sbom.rs
```diff
@@ -0,0 +1,143 @@
...
+            // Make sure that it's possible for this to be a UUID
+            if !report_uuid
+                .chars()
+                .all(|c| c.is_ascii_hexdigit() || c == '-')
+            {
+                return Err(crate::Error::Other {
+                    source: format!("Invalid characters in UUID value '{}'", report_uuid).into(),
+                    backtrace: snafu::Backtrace::capture(),
+                });
+            }
```
XAMPPRocky, line 103: Would it make sense to instead accept a Uuid as the parameter? We could then add a  `fn uuid(&self) -> Uuid` helper method to `crate::models::repos::sbom::SbomGenerateReportResponse`.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `sbom` API for fetching GitHub SBOM information using asynchronous endpoints.
+
```
XAMPPRocky: Changelog is auto generated from your commit message, so you can just format the message to match what you want to include and remove this entry.

XAMPPRocky:
Thank you for your PR! It looks good to me, two minor things stood out.

src/api/repos/sbom.rs
manchicken: I thought about this, but since GitHub replies with a url that some day, in GitHub's infinite wisdom and unilateral breaking-change authority, change to something completely different, it seems to me safer to keep the URL intact and instead to perform a minor security validation in situ.

CHANGELOG.md
manchicken: Ooh, neat! I'll cut this diff then.

manchicken:
I do believe I got the changes addressed. I still think keeping the UUID embedded in the URL is the right call, so I left that one. If you disagree let me know and I can switch it up.

manchicken:
@XAMPPRocky I refreshed the branch this morning, looks like tests and such are still passing. Yay.

XAMPPRocky:
Thank you for your PR!

PR #904, opened 2026-05-26
test: add tests for InMemoryCache [cache-redb 2/3]

ckwalsh:
This commit adds a cache_mem_test.rs integration test, with two test functions. These test cases are implemented generically in a cache_test_helpers.rs file, so they can be reused across cache backends.

While this might be over-engineered for a single test file, it is deliberate. A following PR introduces a ReDB cache implementation, which can then be tested the exact same logic.

This code was written by a human (me) using only "Autocomplete" AI tooling. I provide it to the octocrab project under the terms of the Apache and MIT licenses.

### DIFF
--- a/tests/cache_mem_test.rs
+++ b/tests/cache_mem_test.rs
@@ -1,0 +1,16 @@
+// Tests for caching behavior using InMemoryCache
+mod cache_test_helpers;
+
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR!

PR #906, opened 2026-05-26
feat: redb cache backend [cache-redb 3/3]

ckwalsh:
This commit adds an implementation of the CacheStorage trait backed by
ReDB, "a simple, portable, high-performance, ACID, embedded key-value
store" written in pure Rust.

Data stored in the ReDB database is (de)serialized using Postcard, "a
#![no_std] focused serializer and deserializer for Serde". Postcard
stores data in a binary format and is generally more resource
efficient than json or other text based formats.

The ReDB implementation is gated by the `cache-redb` feature, to avoid
introducing the redb/postcard dependencies for all users.

When the `tracing` feature is enabled, the CacheStorage functions are
annotated with `tracing::instrument`, and any errors reading from /
writing to ReDB are logged with `tracing::error!()` macro.

This code was written by a human (me) using only "Autocomplete" AI
tooling. I provide it to the octocrab project under the terms of the
Apache and MIT licenses.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ request-headers = ["X-GitHub-Api-Version: 2022-11-28"]
 # This list should include all features EXCEPT jwt-aws-lc-rs.
 # When adding new features, remember to add them here too.
 features = [
+    "cache-redb",
     "default-client",
     "follow-redirect",
     "jwt-rust-crypto",
@@ -73,6 +74,8 @@ hyper-tls = { version = "0.6.0", optional = true }
 hyper-util = { version = "0.1.3", features = ["http1"] }
 percent-encoding = "2.2.0"
 pin-project = "1.0.12"
+postcard = { version = "1.1.3", features = ["alloc"], optional = true }
+redb = { version = "4.1.0", optional = true }
 rustls = { version = "0.23", optional = true, default-features = false }
 secrecy = "0.10.3"
 serde = { version = "1.0.126", features = ["derive"] }
@@ -103,6 +106,7 @@ crypto_box = { version = "0.8.2", features = ["seal"] }
 base64 = "0.22.0"
 pretty_assertions = "1.4.0"
 graphql_client = "0.14.0"
+tempfile = "3.27.0"
 
 [build-dependencies]
 cargo_metadata = "0.23.0"
@@ -132,3 +136,8 @@ opentls = ["hyper-tls"]
 stream = ["futures-core", "futures-util"]
 timeout = ["hyper-timeout", "tokio", "tower/timeout"]
 default-client = ["hyper-util/client-legacy"]
+cache-redb = ["redb", "postcard"]
+
+[[test]]
+name = "cache_redb_test"
+required-features = ["cache-redb"]
--- a/src/service/middleware/cache/redb.rs
+++ b/src/service/middleware/cache/redb.rs
@@ -1,0 +1,311 @@
+use std::sync::Arc;
+
+use super::{CacheKey, CacheStorage, CacheWriter, CachedResponse};
+use http::HeaderMap;
+use redb::{Database, ReadableDatabase as _, TableDefinition};
+use serde::{Deserialize, Serialize};
+#[cfg(feature = "tracing")]
+use tracing::instrument;
+
+/// A file-based cache backend using [redb](https://crates.io/crates/redb).
+///
+/// Note, this cache does not implement any eviction policy. Users should be
+/// aware that the cache will grow indefinitely unless the database file is
+/// deleted or truncated.
+#[derive(Debug)]
+pub struct ReDBCache {
+    db: Arc<Database>,
+}
+
+impl ReDBCache {
+    pub fn new<D: Into<Arc<Database>>>(db: D) -> Self {
+        Self { db: db.into() }
+    }
+}
+
+// Wrapper around `http::Uri` to implement `serde::Serialize`,
+// `serde::Deserialize`, `redb::Value` and `redb::Key`.
+#[derive(Debug)]
+enum Uri<'a> {
+    Owned(http::Uri),
+    Ref(&'a http::Uri),
+}
+
+impl Serialize for Uri<'_> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        match self {
+            Uri::Owned(uri) => http_serde::uri::serialize(uri, serializer),
+            Uri::Ref(uri) => http_serde::uri::serialize(uri, serializer),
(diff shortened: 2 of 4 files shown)


ckwalsh:
Was holding this in "draft" for the other two PRs to be merged. Thanks for merging them, this should be ready for review

ckwalsh:
@XAMPPRocky Looks like this PR missed the boat for this weeks release. Anything you want to see to get this merged for the next release?

ckwalsh:
Hey @XAMPPRocky, any blockers for getting this merged? I'd really like to get back to the public release branch for my project

XAMPPRocky:
Thank you for your PR! I just haven't had the time to review it. Cc @maflcko @dmgorsky if you either of you have time to review this addition I would appreciate it :)

maflcko:
Makes sense to disable this by default and only provide it when the option is enabled by the user.

From issue https://github.com/XAMPPRocky/octocrab/issues/175, the preference seems to be to allow users to pick their backend, instead of providing "blessed" ones, so I wonder if redb can be plugged in from the outside instead of being embedded into octocrab?

Personally I haven't yet used caching, because I mostly use webhooks and my polling rarely runs into rate limits. But I can see how other use-cases could benefit from caching.

ckwalsh:
Firstly thank you for the link to that issue. I had not searched for / found previous discussion when putting together this PR.

That's not quite how I read the discussion. I read it as "let's make sure this isn't FS only" rather than "I don't think the core should have cache implementations", but I defer to Rocky to clarify.

I can understand why you might prefer this to be in a separate crate. I think that is possible after the first two PR's, but I'll be honest that that I'm not interested in maintaining a separate crate.

What would you think the next steps are from here?

Cargo.toml
```diff
@@ -28,6 +28,7 @@ features = [
     "jwt-rust-crypto",
     # "jwt-aws-lc-rs",  # Excluded: conflicts with jwt-rust-crypto
     "opentls",
+    "redb",
```
maflcko: nit: Should be `cache-redb`, like below?

(Same in cache.rs)

maflcko (approved):
Ok, I guess if the feature is offered, it is more user-friendly to offer a "batteries-included" experience.

lgtm

Only question is: Should there be a note that there is no LRU/FIFO eviction? Just asking for context, seems fine to implement eviction later, if someone needs it.

ckwalsh:
Good callouts on the review.

Please hold off on merging, I'll rebase and address them tomorrow / this weekend.

ckwalsh:
Apologies for the slow follow up, eviction policy comment added and feature naming fixed

src/service/middleware/cache/redb.rs
```diff
@@ -0,0 +1,311 @@
...
+        Self: 'a,
+    {
+        postcard::from_bytes(data).expect("Failed to deserialize CacheKey from bytes")
+    }
+
+    fn as_bytes<'a, 'b: 'a>(value: &'a Self::SelfType<'b>) -> Self::AsBytes<'a>
+    where
+        Self: 'b,
+    {
+        postcard::to_allocvec(value).expect("Failed to serialize CacheKey")
```
maflcko, line 82:
```suggestion
        postcard::from_bytes(data).expect("Failed to deserialize Uri from bytes")
    }

    fn as_bytes<'a, 'b: 'a>(value: &'a Self::SelfType<'b>) -> Self::AsBytes<'a>
    where
        Self: 'b,
    {
        postcard::to_allocvec(value).expect("Failed to serialize Uri")
```

self-type is uri?

src/service/middleware/cache/redb.rs
```diff
@@ -0,0 +1,311 @@
...
+    fn write_body(&mut self, data: &[u8]) {
+        self.response.body.extend_from_slice(data);
+    }
+}
+
+impl Drop for RedbWriter {
+    #[cfg_attr(feature = "tracing", instrument)]
+    fn drop(&mut self) {
+        // The whole response was received, hence the writer is dropped. We need
+        // to add the response body to the cache.
```
maflcko, line 254: Hmm, if the full body was not received on the drop, this will write partial data?


This is a pre-existing issue, but a possible fix would be:

```diff
diff --git a/src/service/middleware/cache.rs b/src/service/middleware/cache.rs
index 38b9857..e6f41d7 100644
--- a/src/service/middleware/cache.rs
+++ b/src/service/middleware/cache.rs
@@ -59,2 +59,3 @@ pub trait CacheWriter: Send + Sync {
     fn write_body(&mut self, data: &[u8]);
+    fn finish(&mut self);
 }
@@ -222,2 +223,6 @@ where
         match this.inner.poll_frame(cx) {
+            Poll::Ready(None) => {
+                this.writer.finish();
+                Poll::Ready(None)
+            }
             Poll::Ready(frame) => {
diff --git a/src/service/middleware/cache/mem.rs b/src/service/middleware/cache/mem.rs
index 04032b6..6efc006 100644
--- a/src/service/middleware/cache/mem.rs
+++ b/src/service/middleware/cache/mem.rs
@@ -67,8 +67,4 @@ impl CacheWriter for InMemoryWriter {
     }
-}
 
-impl Drop for InMemoryWriter {
-    fn drop(&mut self) {
-        // The whole response was received, hence the writer is dropped. We need
-        // to add the response body to the cache.
+    fn finish(&mut self) {
         let uri = self.uri.clone();
@@ -82 +78,23 @@ impl Drop for InMemoryWriter {
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn dropped_writer_does_not_cache_partial_response() {
+        let cache = InMemoryCache::new();
+        let uri: Uri = "/users/some-user".parse().unwrap();
+
+        let mut writer = cache.writer(
+            &uri,
+            CacheKey::ETag("test-etag".to_owned()),
+            HeaderMap::new(),
+        );
+        writer.write_body(b"partial body");
+        drop(writer);
+
+        assert!(cache.try_hit(&uri).is_none());
+        assert!(cache.load(&uri).is_none());
+    }
+}
diff --git a/src/service/middleware/cache/redb.rs b/src/service/middleware/cache/redb.rs
index eb7d22a..3f8e0be 100644
--- a/src/service/middleware/cache/redb.rs
+++ b/src/service/middleware/cache/redb.rs
@@ -247,9 +247,5 @@ impl CacheWriter for RedbWriter {
     }
-}
 
-impl Drop for RedbWriter {
     #[cfg_attr(feature = "tracing", instrument)]
-    fn drop(&mut self) {
-        // The whole response was received, hence the writer is dropped. We need
-        // to add the response body to the cache.
+    fn finish(&mut self) {

maflcko (approved):
lgtm, just a typo and possible pre-existing issue

PR #918, opened 2026-06-09
Do not serialize input object if it is null (fixes #622)

### LINKED ISSUE #622: create_workflow_dispatch failed with GitHubError
```rust
   octocrab
        .actions()
        .create_workflow_dispatch("owner", "repo", workflow_id, "develop")
        .send()
        .await?;
```

create_workflow_dispatch failed with GitHubError

Error message is `Invalid request.\n\nFor 'properties/inputs', nil is not an object`

andreas-bruckmeier:
https://docs.rs/octocrab/latest/octocrab/actions/struct.ActionsHandler.html#method.create_workflow_dispatch
The call to inputs() is documented as optional:
```
octocrab.actions()
   .create_workflow_dispatch("org", "repo", "workflow.yaml", "ref")
   // optional
   .inputs(serde_json::json!({"my-key": "my-value"}))
   .send()
   .await?;
```
But skipping this call leads to a null-object being sent to Github which leads to a HTTP 422 Invalid Request  with the message "For 'properties/inputs', nil is not an object." This change skips the inputs field from serialization if its null, which is the default.

fixes #622

### DIFF
--- a/src/models/workflows.rs
+++ b/src/models/workflows.rs
@@ -168,5 +168,6 @@ pub struct ArtifactWorkflowRun {
 #[non_exhaustive]
 pub struct WorkflowDispatch {
     pub r#ref: String,
+    #[serde(skip_serializing_if = "serde_json::Value::is_null")]
     pub inputs: serde_json::Value,
 }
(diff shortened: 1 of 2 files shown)


XAMPPRocky:

PR #919, opened 2026-06-09
Use std::sync::LazyLock over once_cell::sync::Lazy

maflcko:
After MSRV 1.85 (commit 174950f0fbdbbbfb2e2a6c38f4f2003831163bca), the `std::` variant of the lazy lock can be used.


XAMPPRocky:
Thank you for your PR!

PR #921, opened 2026-06-14
Document wasm feature for Yew builds

Deivid00007:
## Summary

- Add a `wasm` feature alias for browser-based WebAssembly builds.
- Document the Cargo configuration Yew/browser users need:
  `default-features = false, features = ["wasm"]`.
- Update the existing CI WebAssembly build to use the new feature alias.

## Verification

- `cargo check --target wasm32-unknown-unknown --no-default-features --features wasm`
- `cargo check`

I also confirmed the current default feature set still fails on
`wasm32-unknown-unknown` because native networking dependencies are compiled,
so the README now makes the supported browser/WASM configuration explicit.

## Note

`cargo fmt --all -- --check` could not run on this machine because Windows
Application Control blocked `cargo-fmt` with `os error 4551`. This PR does not
change Rust source files.

Closes XAMPPRocky/octocrab#224.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -65,7 +65,7 @@ jobs:
       - run: |
           rustup default stable
           rustup target add wasm32-unknown-unknown
-          cargo build --target=wasm32-unknown-unknown --no-default-features -F jwt-rust-crypto
+          cargo build --target=wasm32-unknown-unknown --no-default-features -F wasm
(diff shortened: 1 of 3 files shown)


Deivid00007:
Closing this after noticing the maintainer note in #224 that new-contributor PRs are not being accepted for this feature. Sorry for the noise.

PR #924, opened 2026-06-19
fix!: make mandatory PR fields mandatory

flying-sheep:
#902 made fields “optional” that absolutely never are.

This PR reverts that (hopefully saving a few people upgrading from 0.51 to 0.54+ unnecessary churn) and adds some comments that can be used to make more informed decisions in the future.

See #897

### DIFF
--- a/src/models/events/payload/pull_request.rs
+++ b/src/models/events/payload/pull_request.rs
@@ -153,7 +153,7 @@ mod test {
         {
             assert_eq!(payload.action, PullRequestEventAction::Opened);
             assert_eq!(payload.number, 8);
-            assert_eq!(payload.pull_request.id.unwrap().0, 558121796);
+            assert_eq!(payload.pull_request.id.0, 558121796);
         } else {
             panic!("unexpected event payload encountered: {:#?}", event.payload);
         }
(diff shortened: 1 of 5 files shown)


XAMPPRocky (approved):

XAMPPRocky:
Thank you for your PR!

flying-sheep:
I have a follow-up PR but I can’t open one right now due to the restriction. Do you want it?

https://github.com/flying-sheep/octocrab/tree/pr-models

<details>
<summary>PR description</summary>

As promised.

Alternatives: We could use nested structs (flattened by serde) instead, and optionally also traits to have `field` and `field_mut` methods directly on the root struct. But that’s more complex than simple duplication. And even though I implemented `Into`, I doubt many codebases will use them, so their costs are not super relevant.

Maybe TODO:

- [ ] figure out what to do about `user`: it’s mandatory in `SimplePullRequest` and *not* mandatory in `PullRequest` in the schema, otherwise the latter is a straight superset of the former.
- [ ] add new “stack” field (stacked PRs are a new GitHub closed-beta feature)

AI disclaimer: designed by me, code written by Claude, I validated that the fields are in the right bin by checking the JSON schema

</details>

PR #931, opened 2026-07-20
Specify `CryptoProvider` using features

### LINKED ISSUE #930: Could not automatically determine the process-level `CryptoProvider` with default features
Currently, we just specify `hyper-rustls/ring` and `hyper-rustls/aws-lc-rs`, but it don't select provider if features are ambiguous ([rustls docs](https://docs.rs/rustls/latest/src/rustls/crypto/mod.rs.html#261-264)).

We need a way to specify the correct crypto provider, similar to what `reqwest` does: [1](https://github.com/seanmonstar/reqwest/blob/master/src/async_impl/client.rs#L719-L723) + [2](https://github.com/seanmonstar/reqwest/blob/master/src/async_impl/client.rs#L2498-L2511).

Related: #855

Desiders:
Close: #930

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,6 +73,7 @@ hyper-tls = { version = "0.6.0", optional = true }
 hyper-util = { version = "0.1.3", features = ["http1"] }
 percent-encoding = "2.2.0"
 pin-project = "1.0.12"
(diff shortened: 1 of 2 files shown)


XAMPPRocky (approved):

XAMPPRocky:

